Issue #8, opened 2021-08-04
Milestone 3 tasks


rushiagr:
In this milestone, we focus on refactoring changes and make it work with OpenSearch 1.0.0 clusters

- [ ] Namespace changes and other refactoring on *main* branch to support OpenSearch
- [ ] CI is expected to pass for OpenSearch and OpenDistro with and with out security
- [ ] Repo should be made public

PR #47, opened 2021-08-20
Support integ test with security plugin enabled

rushiagr:
Removed a docker-compose.yml file which wasn't being used.

Signed-off-by: Rushi Agrawal <rushi.agr@gmail.com>

### Issues Resolved
Last item on #8 
 
### Check List


test_opensearch/test_server_secured/__init__.py
```diff
@@ -0,0 +1,25 @@
+# SPDX-License-Identifier: Apache-2.0
```
VijayanB, line 1: New file should only have SPDX license.

test_opensearch/test_server_secured/test_clients.py
```diff
@@ -0,0 +1,48 @@
...
+            verify_certs=False,
+        )
+
+        info = client.info()
+        self.assertNotEqual(info["version"]["number"], "")
+        self.assertNotEqual(info["tagline"], "")
+        self.assertTrue(
+            info["version"].get("build_flavor", "") != ""
+            or info["version"].get("distribution", "") != ""
+        )
```
axeoman, line 48: Have no idea what OpenSearch build with different options can return, but does that mean that keys inside `version` can be an empty string which is a criterion of failure? (Trying to understand, is there any motivation to not use e.g. `"build_flavor" in info["version"]` syntax or not).

axeoman (commented):
LGTM. One typo and one detail asked from curiosity.

test_opensearch/test_server_secured/test_clients.py
VijayanB: i agree, ```"build_flavor" in info["version"]``` is more pythonic than using !=

erickguan:
When the client specification will be public?

test_opensearch/test_server_secured/__init__.py
rushiagr: Addressed in latest commit

test_opensearch/run_tests.py
```diff
@@ -113,18 +117,34 @@ def run_all(argv=None):
...
                     "test_opensearch/test_server/",
+                    "test_opensearch/test_server_secured/",
                     "test_opensearch/test_async/test_server/",
                 ]
             )
-        if ignores:
-            argv.extend(["--ignore=%s" % ignore for ignore in ignores])
 
-        # Jenkins, only run server tests
+        # Jenkins/gihub actions, only run server tests
```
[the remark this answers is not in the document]
rushiagr: Addressed in latest commit

test_opensearch/test_server_secured/test_clients.py
rushiagr: Addressed in latest commit

.ci/run-opensearch.sh
```diff
@@ -65,15 +65,19 @@ END
...
     --file=.ci/$CLUSTER/Dockerfile \
-    --build-arg SECURE_INTEGRATION=false \
-    --tag=$CLUSTER \
+    --build-arg SECURE_INTEGRATION=$SECURE_INTEGRATION \
+    --tag=$CLUSTER-secure-$SECURE_INTEGRATION \
     .
 
   echo -e "\033[34;1mINFO:\033[0m Starting container $node_name \033[0m"
   set -x
+  healthcmd="curl -vvv -s --fail localhost:9200/_cluster/health || exit 1"
```
deztructor, line 76: I wonder wouldn't explicit `http://` protocol improve reading experience.

.ci/run-tests
```diff
@@ -8,13 +8,19 @@
 export TEST_SUITE="${TEST_SUITE:=oss}"
 export PYTHON_VERSION="${PYTHON_VERSION:=3.9}"
 export PYTHON_CONNECTION_CLASS="${PYTHON_CONNECTION_CLASS:=Urllib3HttpConnection}"
-export OPENSEARCH_URL_EXTENSION="${OPENSEARCH_URL_EXTENSION:=http}"
-export CLUSTER="${1:=opensearch}"
+export CLUSTER="${1:-opensearch}"
+export SECURE_INTEGRATION="${2:-false}"
+if [[ "$SECURE_INTEGRATION" == "true" ]]; then
```
deztructor, line 13: Could be simplified if you use it as a boolean expression:
```
if "$SECURE_INTEGRATION"; then
...
```

.ci/run-opensearch.sh
```diff
@@ -65,15 +65,19 @@ END
...
+  if [[ "$SECURE_INTEGRATION" == "true" ]]; then
```
deztructor, line 77: Could be simplified if you use it as a boolean expression:
```
if "$SECURE_INTEGRATION"; then
...
```

test_opensearch/run_tests.py
```diff
@@ -113,18 +117,34 @@ def run_all(argv=None):
             ignores.extend(
                 [
                     "test_opensearch/test_server/",
+                    "test_opensearch/test_server_secured/",
```
deztructor, line 120: It seems it should be added to ignores only for the case when `not secured` -> done on line 140.

deztructor: Ah, I see, it is for `if "GITHUB_ACTION" in environ:`

.ci/run-tests
rushiagr: Not an expert in bash, but looks like it's preferable to do comparison the way I've done, since it's a string which we get as an input to the `run-tests` file.

.ci/run-opensearch.sh
rushiagr: Same as above

.ci/run-opensearch.sh
rushiagr: Updated in latest revision

rushiagr:
> When the client specification will be public?

Hi @erickguan , do you mean the documentation, or something else?

PR #53, opened 2021-08-24
Add condition for shallow clone 

### LINKED ISSUE #52: [BUG] Integ test does a deep checkout of opensearch repository
**What is the bug?**
[Here](https://github.com/opensearch-project/opensearch-py/blob/340069f6fa344ab3853fec8c48e60e6a7756f271/test_opensearch/run_tests.py#L49), we're doing a deep checkout of OpenSearch (the server) repository. This works reasonably fast on Github Actions, but slows down the run time considerably when doing it on a developer laptop. We should ideally do a shallow checkout i.e. `git clone --depth=1 <url>`

Run `./.ci/run-test opensearch true` on dev laptop (requires Docker to be installed)

The build should finish fast

 - OS: MacOS Catalina

N/A

N/A

vchrombie:
### Description

The integration tests perform a deep checkout of the opensearch repository. Instead, we can perform a shallow clone the main branch to improve the speed and performance of the tests.
 
### Issues Resolved

Fixes #52 
 
### Check List

### DIFF
--- a/test_opensearch/run_tests.py
+++ b/test_opensearch/run_tests.py
@@ -46,7 +46,8 @@ def fetch_opensearch_repo():
     # no repo
     if not exists(repo_path) or not exists(join(repo_path, ".git")):
         subprocess.check_call(
-            "git clone https://github.com/opensearch-project/opensearch %s" % repo_path,
+            "git clone --branch main --single-branch --depth 1 "
+            "https://github.com/opensearch-project/opensearch %s" % repo_path,
             shell=True,
         )
 
@@ -62,7 +63,7 @@ def fetch_opensearch_repo():
     from test_opensearch.test_cases import SkipTest
     from test_opensearch.test_server import get_client
 
-    # find out the sha of the running client
+    # find out the sha of the running server
     try:
         client = get_client()
         sha = client.info()["version"]["build_hash"]
@@ -77,9 +78,12 @@ def fetch_opensearch_repo():
         % repo_path,
         shell=True,
     )
+
     # reset to the version from info()
-    subprocess.check_call("cd %s && git fetch" % repo_path, shell=True)
-    subprocess.check_call("cd %s && git reset --hard %s" % (repo_path, sha), shell=True)
+    subprocess.check_call(
+        "cd %s && git fetch --depth=1 origin %s" % (repo_path, sha),
+        shell=True,
+    )
 
 
 def run_all(argv=None):


vchrombie:
The build fails. 
https://github.com/vchrombie/opensearch-py/runs/3415891606

> subprocess.CalledProcessError: Command 'cd /code/opensearch && git reset --hard 34550c5b17124ddc59458ef774f6b43a086522e3' returned non-zero exit status 128.

Since, we are not cloning the whole repository, it couldn't find the `34550c5b17124ddc59458ef774f6b43a086522e3` commit I guess. We need to find a workaround for this.

As a hotfix, I commented out the line and the build passes.
https://github.com/vchrombie/opensearch-py/runs/3415915005
```py
    print(sha)
    # subprocess.check_call("cd %s && git reset --hard %s" % (repo_path, sha), shell=True)
```

vchrombie:
I just pushed a commit https://github.com/vchrombie/opensearch-py/commit/cab41f49bbc974c6fc5b3c93171d63cf0f39866f to a new branch `test-issue-52`. I tried removing the code from [run_tests.py#L61](https://github.com/opensearch-project/opensearch-py/blob/main/test_opensearch/run_tests.py#L61) to [run_tests.py#L82](https://github.com/opensearch-project/opensearch-py/blob/main/test_opensearch/run_tests.py#L82). Not sure if it is the right thing to do, but the workflows seemed to be almost the same.

- https://github.com/vchrombie/opensearch-py/runs/3418310013
- https://github.com/vchrombie/opensearch-py/runs/3415915005

I was just skeptical because we are missing out on this condition if opensearch is running [run_tests.py#L70](https://github.com/opensearch-project/opensearch-py/blob/main/test_opensearch/run_tests.py#L70).

I would like to listen to your comments on this.

rushiagr:
If you check here https://github.com/opensearch-project/opensearch-py/blob/6fcc78a58cc84396a5b771d829c38f28623cce1a/test_opensearch/run_tests.py#L68  the hash is derived by calling the server code, and we need to check out at that point. I think it is important that we pull server code at the correct commit, else it'll have problems in future when there are multiple releases/branches. 

Taking inspiration from [this](https://stackoverflow.com/a/3489576/1143173) stackoverflow, we can do something like this:

```
mkdir ~/tmp
cd ~/tmp
git remote add origin git@github.com:opensearch-project/opensearch-py.git
git fetch --depth=1 origin $SHA # sha is the one derived from build_hash
# do work in this directory
rm -rf ~/tmp
```

Sorry if this increases the scope of the work in this PR :)

Also, while you're at it, if possible, could you also change this line
https://github.com/opensearch-project/opensearch-py/blob/6fcc78a58cc84396a5b771d829c38f28623cce1a/test_opensearch/run_tests.py#L65
to say 'server' instead of 'client'? Thanks again.

vchrombie:
Hi @rushiagr, I completed the work as per your suggestions.

Can you take a look when you are free?

Looks like the time is reduced to almost half in github actions. I haven't tested locally tho.
https://github.com/opensearch-project/opensearch-py/actions/runs/1187109252 took 5m 6s 
https://github.com/vchrombie/opensearch-py/actions/runs/1189548548 took  2m 44s

test_opensearch/run_tests.py
```diff
@@ -46,7 +46,8 @@ def fetch_opensearch_repo():
     # no repo
     if not exists(repo_path) or not exists(join(repo_path, ".git")):
         subprocess.check_call(
-            "git clone https://github.com/opensearch-project/opensearch %s" % repo_path,
+            "git clone --branch main --single-branch --depth 1 "
```
rushiagr, line 49: I think the problem with this approach would be, that once we have multiple versions of OpenSearch server (and clients too), the branch 'main' would not contain the requisite commit. For example, say the 'main' is OpenSearch version 3.x, and the OpenSearch server against which we are testing is on version 1.0, then the build hash returned by server will be a commit from '1.0' branch, and might not be present on the 'main' branch..

vchrombie: Hmm, that's interesting.

Do you think removing the `--branch main --single-branch` would work?
```py
            "git clone --depth 1 https://github.com/opensearch-project/opensearch %s" % repo_path,
```

Thanks for the review.

rushiagr: That still might not work, as the required commit might not be present on the main branch (branch 'main' will get picked up by default). I suggested a method in an earlier comment, https://github.com/opensearch-project/opensearch-py/pull/53#issuecomment-905407817, can you try that, or something similar?

vchrombie: I think I got it right this time.
@rushiagr, can you please check #70 when you have time.

vchrombie:
Closing this PR in favour of https://github.com/opensearch-project/opensearch-py/pull/70.

PR #76, opened 2021-09-21
Update README: use pypi instead of git repo for installation

rushiagr:
### Description
Update README: use pypi instead of git repo for installation
 
### Issues Resolved
None
 
### Check List

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ This is the low-level client. A high-level Python client is in the works, and wi
 To add the client to your project, install it using [pip](https://pip.pypa.io/):
 
 ```bash
-pip install git+https://github.com/opensearch-project/opensearch-py.git
+pip install opensearch-py
 ```
 
 Then import it like any other module:


vchrombie:
I think the pypi badge can also be added.

[PyPI version](https://badge.fury.io/py/opensearch-py)

```
[![PyPI version](https://badge.fury.io/py/opensearch-py.svg)](https://badge.fury.io/py/opensearch-py)
```

README.md
```diff
@@ -26,7 +26,7 @@ This is the low-level client. A high-level Python client is in the works, and wi
 To add the client to your project, install it using [pip](https://pip.pypa.io/):
 
 ```bash
-pip install git+https://github.com/opensearch-project/opensearch-py.git
+pip install opensearch-py
```
vamshin, line 29: Thinking out loud here. Should we provide both options?

Question: If we want to download specific client version say 1.0 or 1.1, how can we do this?

vchrombie: > Thinking out loud here. Should we provide both options?

Looks like a good idea. How about the below steps.

---
## Source Code
```
$ git clone https://github.com/opensearch-project/opensearch-py
$ cd opensearch-py
$ python setup.py install
```

## PyPI
```
$ pip install opensearch-py
```
---

`python setup.py install` can be replaced with `pip install .` too I guess (might need `wheel` package as well).

> If we want to download specific client version say 1.0 or 1.1, how can we do this? f this is possible, then lets also add this to README?

We can do this in both the ways again.

---
## Source Code
```
$ git clone https://github.com/opensearch-project/opensearch-py
$ cd opensearch-py
$ git checkout tags/1.1.0 -b 1.1.0
$ python setup.py install
```

## PyPI
```
$ pip install opensearch-py==1.1.0
```
---

We usually provide different option in our tools like [chaoss/grimoirelab-perceval - #installation](https://github.com/chaoss/grimoirelab-perceval#installation).

Thanks.

vamshin: Looks Great. Thanks @vchrombie 

Having Github and Pip install options part of 2 sections like below seems helpful to me.

1. Download latest version
      - Source code
      - pip  

2. Download specific version
      - Source code
      - pip

VijayanB:
Created ticket to handle feedback here [https://github.com/opensearch-project/opensearch-py/issues/79](url)

VijayanB (approved):

PR #91, opened 2021-11-16
Parse raw_data on error only if the Content-Type is application/json.

mccullocht:
### Description
Some opensearch REST APIs will return a Content-Type of text/plain for certain
status codes which will trigger logging when the response is not parsable as
json. Checking the content-type first avoids this error.

This frequently happens with 401s, which occurs when we rotate user keys.
The input that triggers this behavior can be triggered with `curl`:

```
$  curl -k -u"admin:wrong" -w"\n%{http_code}\n%{content_type}\n" https://localhost:9200/_cluster/health
Unauthorized
401
text/plain; charset=UTF-8
```
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-py/issues/92
 
### Check List
 
Signed-off-by: Trevor McCulloch <trevor.mcculloch@benchling.com>


opensearchpy/connection/http_urllib3.py
```diff
@@ -279,7 +279,11 @@ def perform_request(
             self.log_request_fail(
                 method, full_url, url, orig_body, duration, response.status, raw_data
             )
-            self._raise_error(response.status, raw_data)
+            self._raise_error(
+                response.status,
+                raw_data,
+                response.headers.get("Content-Type", None),
```
VijayanB, line 285: create a new method to get this values as
```     
     def get_response_headers(self, response)
          response_headers = {
                header.lower(): value for header, value in response.headers.items()
            }
```
and later
```
  self._raise_error(
      response.status,
      raw_data,
      self.get_response_headers(.get("content-type", "")),
```

opensearchpy/connection/base.py
```diff
@@ -314,12 +315,17 @@ def log_request_fail(
...
         """Locate appropriate exception and raise it."""
         error_message = raw_data
         additional_info = None
         try:
-            if raw_data:
+            content_type = (
+                cgi.parse_header(content_type)[0]
+                if content_type
+                else "application/json"
+            )
```
VijayanB, line 327: shall we pass value (eg: application/json or text ) and avoid this parser

mccullocht: Replaced with something cruder (we only need the first part of the mime type)

opensearchpy/connection/http_urllib3.py
mccullocht: Done

axeoman (changes requested):
Thanks for the contribution and apologies for the delay. 
Overall looks good, could you please [fix](https://github.com/opensearch-project/opensearch-py/runs/4585804827?check_suite_focus=true) the formatting by running black locally and [sign](https://github.com/opensearch-project/opensearch-py/runs/4585804841?check_suite_focus=true) the commits. 
Also, rebase on fresh master will help with passing integration tests.

mccullocht:
Rebased and signed the commits. Fixed formatting error http_urllib3.py. PTAL

opensearchpy/connection/base.pyi
```diff
@@ -112,6 +112,8 @@ class Connection(object):
         response: Optional[str] = ...,
         exception: Optional[Exception] = ...,
     ) -> None: ...
-    def _raise_error(self, status_code: int, raw_data: str) -> NoReturn: ...
+    def _raise_error(
+        self, status_code: int, raw_data: str, content_type: Optional[str]
```
axeoman, line 116: Sorry, did not notice it, the correct type of `content_type` should be `str` since passing `None` will lead to an `AttributeError`. 
To be honest, I would prefer having `Optional`, because it is better than passing empty strings if there is no `Content-Type` available, but we need to have a check before trying to `.split(";")` in that case.

mccullocht: Changed the implementation to be `Optional[str]` with default value of `None`

axeoman: Thanks!
Not sure if we want to parse JSON if there is no "Content-Type" header. Also `dict.get(value)` is sufficient (None is returned by default if key is missing).

mccullocht: Now defaults to text/plain. removed default argument to `dict.get(k)`

axeoman: Thanks. Tests are [failing](https://github.com/opensearch-project/opensearch-py/runs/4605453962?check_suite_focus=true) with `E   AssertionError: no logs of level INFO or higher triggered on opensearch` - could you please take a look?

mccullocht: Fixed. The test assumed content would be treated as application/json by default, which is no longer true.

axeoman (approved):

PR #99, opened 2021-12-15
Use 1.0.0 instead of latest tag

### LINKED ISSUE #97: [BUG] CI fails for main
**What is the bug?**
CI fails for integration test ( opensearch, true )

https://github.com/opensearch-project/opensearch-py/runs/4186359419?check_suite_focus=true

CI should pass

Github

N/A

We should not run against latest since it is a moving target. We should run against released tested version and increment only after we verify it works with this library.

VijayanB:
Signed-off-by: Vijayan Balasubramanian <balasvij@amazon.com>

### Description
Add default value as 1.0.0.
 
### Issues Resolved
#97 
 
### Check List

### DIFF
--- a/.ci/opensearch/Dockerfile
+++ b/.ci/opensearch/Dockerfile
@@ -1,5 +1,5 @@
-ARG OPENSEARCH_VERSION
-FROM opensearchproject/opensearch:latest
+ARG OPENSEARCH_VERSION=1.0.0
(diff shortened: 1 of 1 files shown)


axeoman (approved):
Thanks, I was also thinking about using 1.0.0, but I noticed that starting from version 1.1.0 lacks "kirk" certificates inside OpenSearch container and that is why the error is happening.

Issue #85, opened 2021-10-21
[FEATURE] Add IAM authentication option


aetter:
If people want to connect to an Amazon OpenSearch Service cluster, they have to figure out how to use this client alongside some third-party signing library and/or the AWS SDK, which is non-trivial due to how we (by design) conceal the underlying HTTP requests that the client makes. We should offer IAM signing as an option, just like the OpenSearch CLI does.

**Is your feature request related to a problem?**

If you want to use this client with an Amazon OpenSearch Service cluster that has IAM authentication rather than basic authentication, it's a challenge. The best solution is to use the `requests_aws4auth` library.

**What solution would you like?**

Something like how `requests_aws4auth` works:

```
awsauth = AWS4Auth(credentials.access_key, credentials.secret_key, region, service, session_token=credentials.token)

search = OpenSearch(
    hosts = [{'host': host, 'port': 443}],
    http_auth = awsauth,
    use_ssl = True,
    verify_certs = True,
    connection_class = RequestsHttpConnection
)
```

**What alternatives have you considered?**

Maintaining the status quo, which is admittedly the best of all the clients, but still suboptimal.

**Do you have any additional context?**

https://github.com/opensearch-project/OpenSearch/issues/1400

Shivamdhar:
Released this feature with opensearch-py `1.1.0`

lizsnyder:
@Shivamdhar can you provide sample code?

Shivamdhar:
Please refer this section for the sample usage of AWS IAM auth: https://github.com/opensearch-project/opensearch-py#using-iam-credentials-for-authentication

lizsnyder:
Perfect, thanks @Shivamdhar. I'll add a similar sample to the AWS docs.

harshavamsi:
> Please refer this section for the sample usage of AWS IAM auth: https://github.com/opensearch-project/opensearch-py#using-iam-credentials-for-authentication

Newer link: https://github.com/opensearch-project/opensearch-py/blob/main/GETTING_STARTED.md#using-iam-credentials-for-authentication

GuanLinZhang:
Link has been updated to https://github.com/opensearch-project/opensearch-py/blob/main/USER_GUIDE.md#using-iam-credentials-for-authentication

PR #133, opened 2022-02-17
adds AwsV4Signer implementation in the library

Shivamdhar:
Signed-off-by: Shivam Dhar <dhshivam@amazon.com>

### Description
Incorporate AWS IAM signing support in the library
 
### Issues Resolved
https://github.com/opensearch-project/opensearch-py/issues/85

### Usage

```
from opensearchpy import OpenSearch, RequestsHttpConnection, AWSV4Signer
import boto3

host = ''
region = 'us-west-2' # e.g. us-west-1
service = 'es'
auth = AWSV4Signer(credentials, region)

search = OpenSearch(
    hosts = [{'host': host, 'port': 443}],
    http_auth = auth,
    use_ssl = True,
    verify_certs = True,
    connection_class = RequestsHttpConnection
)
```
 
**Requirement**: _The consumers who want to make use of AWSV4Signer, should have python version 3.6 or above._

### Check List

### DIFF
--- a/opensearchpy/helpers/signer.py
+++ b/opensearchpy/helpers/signer.py
@@ -1,0 +1,128 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+#
+# Modifications Copyright OpenSearch Contributors. See
+# GitHub history for details.
+#
+#  Licensed to Elasticsearch B.V. under one or more contributor
+#  license agreements. See the NOTICE file distributed with
+#  this work for additional information regarding copyright
+#  ownership. Elasticsearch B.V. licenses this file to you under
+#  the Apache License, Version 2.0 (the "License"); you may
+#  not use this file except in compliance with the License.
+#  You may obtain a copy of the License at
+#
+# 	http://www.apache.org/licenses/LICENSE-2.0
+#
+#  Unless required by applicable law or agreed to in writing,
+#  software distributed under the License is distributed on an
+#  "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+#  KIND, either express or implied.  See the License for the
+#  specific language governing permissions and limitations
+#  under the License.
+
+import sys
+
+import requests
+
+from ..exceptions import ImproperlyConfigured
+
+OPENSEARCH_SERVICE = "es"
+
+
+def python_version_gte_36():  # type: ignore
+    """
+    A util method to check python version.
+    :return: boolean that returns true if python version running is greater than
+    equal to 3.6 otherwise returns false
+    """
+    if sys.hexversion >= 0x03060000:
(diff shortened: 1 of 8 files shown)


opensearchpy/helpers/signer.py
```diff
@@ -0,0 +1,128 @@
...
+        return False
+
+
+try:
+    if python_version_gte_36():  # type: ignore
+        from urllib.parse import parse_qs, urlencode, urlparse
+except ImportError:
+    raise ImproperlyConfigured(
+        "Please upgrade the python version to 3.6 or above to use AWS Signer."
+    )
```
dblock, line 54: Should the entire client require 3.6 or newer?

opensearchpy/helpers/signer.py
```diff
@@ -0,0 +1,128 @@
...
+            SigV4Auth(self.credentials, OPENSEARCH_SERVICE, self.region).add_auth(
+                aws_request
+            )
+
+            # copy the headers from AWS request object into the prepared_request
+            prepared_request.headers.update(dict(aws_request.headers.items()))
+
+            return prepared_request
+        except ImportError:
+            raise ImproperlyConfigured("Please install botocore to use AwsV4Signer.")
```
dblock, line 103: Shouldn't this fail the same way as for Python 3.6 on load and not when actually making a request?

dblock (commented):
Needs more tests (e.g. when missing region) and documentation of how to use this, please!

opensearchpy/helpers/signer.py
VijayanB: At the moment, 1.x version contains support for both python 2.x and 3.x. Unfortunately, SigV4 from botocore recommends only python 3.x support. If we want to drop python 2.x support, we can do it only in next major version change like opensearch-py 2.0.
I see two options,
1. Create 2.x branch, drop python 2.x support and include this feature only in 2.x
2. Have this feature in 1.x with catch import error like Django.

opensearchpy/helpers/signer.py
```diff
@@ -0,0 +1,128 @@
...
+import sys
+
+import requests
+
+from ..exceptions import ImproperlyConfigured
+
+OPENSEARCH_SERVICE = "es"
+
+
+def python_version_gte_36():  # type: ignore
```
VijayanB, line 36: IMO, we don't have to check for python version. Instead, just catch for import error when someone try to use SigV4 and fail with `ImproperlyConfigured` that it is not supported.

opensearchpy/helpers/signer.py
Shivamdhar: Option 1 sounds good to me @VijayanB

Shivamdhar:
> Needs more tests (e.g. when missing region) and documentation of how to use this, please!

We will add more tests. The PR has the usage section, we will add it to the README once the PR is merged.

Shivamdhar:
Closing this PR in favor of https://github.com/opensearch-project/opensearch-py/pull/136

PR #210, opened 2022-09-29
Added timeout to available input query parameters

### LINKED ISSUE #209: [BUG] _resolve_request_args looking for "request_timeout" instead of "timeout"
In the documentation for bulk(), the key "timeout" is listed as an option to send to the params input, however, _resolve_request_args() seems to be looking for the param "request_timeout" instead of "timeout". The key "timeout" doesn't appear to change the timeout, but "request_timeout" does appear to work.

For example:

This appears to change the timeout
```python
response = client.bulk(documents, params={"request_timeout":1000000})
```

But this doesn't appear to change the timeout
```python
response = client.bulk(documents, params={"timeout":1000000})
```

But the documentation specifies to use the key "timeout".

bulk docstring (line 394) in opensearchpy/client/__init__.py
```python
def bulk(self, body, index=None, params=None, headers=None):
        """
        Allows to perform multiple index/update/delete operations in a single request.


        :arg body: The operation definition and data (action-data
            pairs), separated by newlines
        :arg index: Default index for items which don't provide one
        :arg _source: True or false to return the _source field or not,
            or default list of fields to return, can be overridden on each sub-
            request
        :arg _source_excludes: Default list of fields to exclude from
            the returned _source field, can be overridden on each sub-request
        :arg _source_includes: Default list of fields to extract and
            return from the _source field, can be overridden on each sub-request
        :arg pipeline: The pipeline id to preprocess incoming documents
            with
        :arg refresh: If `true` then refresh the affected shards to make
            this operation visible to search, if `wait_for` then wait for a refresh
            to make this operation visible to search, if `false` (the default) then
            do nothing with refreshes.  Valid choices: true, false, wait_for
        :arg require_alias: Sets require_alias for all incoming
            documents. Defaults to unset (false)
        :arg routing: Specific routing value
        :arg timeout: Explicit operation timeout
        :arg wait_for_active_shards: Sets the number of shard copies
            that must be active before proceeding with the bulk operation. Defaults
            to 1, meaning the primary shard only. Set to `all` for all shard copies,
            otherwise set to any non-negative value less than or equal to the total
            number of copies for the shard (number of replicas + 1)
```

_resolve_request_args (line 456) in opensearchpy/transport.py
```python
timeout = None
if params:
    timeout = params.pop("request_timeout", None)
    ignore = params.pop("ignore", ())
    if isinstance(ignore, int):
        ignore = (ignore,)

        return method, params, body, ignore, timeout
```

harshavamsi:
Signed-off-by: Harsha Vamsi Kalluri <harshavamsi096@gmail.com>

### Description
Fixes the timeout query parameter to allow for not just `request_timeout` but also `timeout` like the examples say.

### Issues Resolved
Fixes #209 

### DIFF
--- a/opensearchpy/transport.py
+++ b/opensearchpy/transport.py
@@ -454,6 +454,8 @@ class Transport(object):
         timeout = None
         if params:
             timeout = params.pop("request_timeout", None)
+            if not timeout:
+                timeout = params.pop("timeout", None)
             ignore = params.pop("ignore", ())
             if isinstance(ignore, int):
                 ignore = (ignore,)
--- a/test_opensearchpy/test_async/test_transport.py
+++ b/test_opensearchpy/test_async/test_transport.py
@@ -139,6 +139,18 @@ class TestTransport:
             "headers": None,
         } == t.get_connection().calls[0][1]
 
+    async def test_timeout_extracted_from_params_and_passed(self):
+        t = AsyncTransport([{}], connection_class=DummyConnection)
+
+        await t.perform_request("GET", "/", params={"timeout": 84})
+        assert 1 == len(t.get_connection().calls)
+        assert ("GET", "/", {}, None) == t.get_connection().calls[0][0]
+        assert {
+            "timeout": 84,
+            "ignore": (),
+            "headers": None,
+        } == t.get_connection().calls[0][1]
+
     async def test_opaque_id(self):
         t = AsyncTransport([{}], opaque_id="app-1", connection_class=DummyConnection)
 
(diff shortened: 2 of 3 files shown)


dblock:
Needs a unit test, please?

harshavamsi:
@dblock added.

dblock (approved):
Why 84 and not 42?! :)

dblock (changes requested):
~One more thing, it needs to be documented in user guide/README.~

I realize we don't have this here, or at least it doesn't look like other clients that have a USERS_GUIDE? @VachaShah what are we to do for this one?

VachaShah:
> 

USER_GUIDE is the [GETTING_STARTED](https://github.com/opensearch-project/opensearch-py/blob/main/GETTING_STARTED.md) guide. @harshavamsi If you are making the change to this guide in your PR, can you also rename it to USER_GUIDE?

axeoman (approved):

PR #214, opened 2022-10-06
Add kerberos support

psaiz:
### Description
Provide kerberos authentication support to connect to clusters

### Issues Resolved
This pull request creates a new connection class that extends the HTTP connection to support kerberos. Thanks to it, using the code below, the user can connect using kerberos for authentication:

```
from opensearchpy import OpenSearch
from opensearchpy.connection.krb_requests  import RequestsKrbConnection

client = OpenSearch(["https://<host>/"], connection_class=RequestsKrbConnection)
```


harshavamsi:
Hi @psaiz, thanks for the contribution. A couple of thoughts.

We need to get DCO check to pass, so could you sign off your commit with the `-s` flag.

I noticed you're importing `kerberos`, I believe we would need to include it as a dependency to be installed in `setup.py`

We would need unit tests to make sure that this works for both sync and async connections. This ensures test compatibility. 

And it would be great if you could add this class to the `docs` folder so that we can auto generate docs.

psaiz:
Hi @harshavamsi, 
Thanks for your comments. I'll do the docs and I'll look into the tests and the -s flag, thanks!

About adding kerberos as a dependency, what would you think about making it an [optional dependency](https://setuptools.pypa.io/en/latest/userguide/dependency_management.html#optional-dependencies)? I mean, the opensearch-py can still be used without kerberos. 
Thanks again ,

Pablo

dblock:
> Regarding the testing, I'm not sure how to proceed. Note that to test it properly it requires a kerberos setup, and I'm afraid that won't be available during the pipelines. I've checked how it is done in the [pykerberos](https://github.com/bgamble/pykerberos), and they give a test.py file that should be run against their system. Is that what you have in mind?

At least exercise the code to _some_ extent. Mock the kerberos setup.

psaiz:
Hi @dblock, 
Thanks for the message. Here you have a very basic unittest mocking the kerberos calls.  Let me know if this is what you had in mind.

Best, 
pablo

opensearchpy/connection/krb_requests.py
```diff
@@ -0,0 +1,64 @@
+"""
+Class for the kerberos authentication.
```
dblock, line 2: Nit: "Implements a connection with Kerberos authentication support." (cause we know it's a class ;))

opensearchpy/connection/krb_requests.py
```diff
@@ -0,0 +1,64 @@
...
+During the connection phase, one header is added with the kerberos context
```
dblock, line 3: Adds an Authorization header during ...

opensearchpy/connection/krb_requests.py
```diff
@@ -0,0 +1,64 @@
...
+                 verify_certs=True,
+                 ssl_show_warn=True,
+                 ca_certs=None,
+                 client_cert=None,
+                 client_key=None,
+                 headers=None,
+                 http_compress=None,
+                 cloud_id=None,
+                 api_key=None,
+                 opaque_id=None,
```
dblock, line 26: Why is set of arguments different from the super class? Aren't we introducing defaults where we shouldn't?

opensearchpy/connection/krb_requests.py
```diff
@@ -0,0 +1,64 @@
...
+        """do the kerberos magic"""
+        principal = 'HTTP@' + host
+        try:
+            dummy, _krb_context = kerberos.authGSSClientInit(principal)
+        except Exception as message: # pylint: disable=broad-except
+            raise ValueError(f"krb_contex error: {message}")
+        try:
+            kerberos.authGSSClientStep(_krb_context, "")
+        except kerberos.GSSError as message:
+            warnings.warn(f"Cannot connect to the realm {principal}")
```
dblock, line 54: Include message here too in the `warn`?

opensearchpy/connection/krb_requests.py
```diff
@@ -0,0 +1,64 @@
...
+        )
+        self._set_kerberos_headers(host)
+
+
+    def _set_kerberos_headers(self, host):
+        """do the kerberos magic"""
+        principal = 'HTTP@' + host
+        try:
+            dummy, _krb_context = kerberos.authGSSClientInit(principal)
+        except Exception as message: # pylint: disable=broad-except
```
dblock, line 49: 1. Why do we swallow exceptions here to raise something more generic?
2. Is there a way to preserve the inner exception in the re-raised exception?

test_opensearchpy/test_krbconnection.py
```diff
@@ -0,0 +1,19 @@
...
+
+class TestKrbConnection(unittest.TestCase): # pylint: disable=invalid-name
+    """ Test for the RequestsKrbConnection """
+
+    @patch('kerberos.authGSSClientStep')
+    @patch('kerberos.authGSSClientResponse')
+    def test_krbConnection(self, authGSSClientResport, authGSSClientStep): # pylint: disable=unused-argument,invalid-name
+        """ Check if we can we a Kerberos connection, mocking up the kerberos calls """
+        my_object = RequestsKrbConnection()
+        self.assertTrue('Authorization' in my_object.session.headers)
```
dblock, line 16: Good start. Mock a little further, ensure that the `kerberos.` methods are called, and add a check for the value, it should be "Negotiate [dummy token returned from a mock]" or something like that.

dblock (changes requested):
Close!

I made some other comments a bit more in-depth.

opensearchpy/connection/krb_requests.py
```diff
@@ -0,0 +1,62 @@
...
+import kerberos
+
+from . import RequestsHttpConnection
+
+class RequestsKrbConnection(RequestsHttpConnection):
+    """ Class to connect with Kerberos """
+    def __init__(self,
+                 host="localhost",
+                 port=None,
+                 http_auth=None,
```
dblock, line 15: This looks unused, what's `http_auth`?

opensearchpy/connection/krb_requests.py
```diff
@@ -0,0 +1,62 @@
...
+            headers=headers,
+            http_compress=http_compress,
+            opaque_id=opaque_id,
+            **kwargs
+        )
+        self._set_kerberos_headers(host)
+
+
+    def _set_kerberos_headers(self, host):
+        """do the kerberos magic"""
```
dblock, line 46: Let's write a useful comment or remove ;) It's no magic.

opensearchpy/connection/krb_requests.py
```diff
@@ -0,0 +1,62 @@
...
+        principal = 'HTTP@' + host
+        _, _krb_context = kerberos.authGSSClientInit(principal)
+        try:
+            kerberos.authGSSClientStep(_krb_context, "")
+        except kerberos.GSSError as message:
+            warnings.warn(f"Cannot connect to the realm {principal}. More details: {message}")
```
dblock, line 52: Showing client-side warnings isn't a pattern in opensearch-py, I couldn't find another place where this is done. I suppose `GSSError` contains enough information about the problem, just let it be? And what is the reason to wrap all exceptions into `ValueError`?

opensearchpy/connection/krb_requests.py
```diff
@@ -0,0 +1,62 @@
...
+            raise message
+        except Exception as message: # pylint: disable=broad-except
+            raise ValueError(f"krb GSSClientStep error: {message}")
+
+        negotiate_details = kerberos.authGSSClientResponse(_krb_context)
+
+        if not negotiate_details:
+            warnings.warn(f"Could not get the kerberos connection details")
```
dblock, line 60: Remove the warning.

test_opensearchpy/test_krbconnection.py
```diff
@@ -0,0 +1,30 @@
...
+class TestKrbConnection(unittest.TestCase): # pylint: disable=invalid-name
+    """ Test for the RequestsKrbConnection """
+
+    @patch('kerberos.authGSSClientStep')
+    @patch('kerberos.authGSSClientResponse')
+    def test_krbConnection(self, authGSSClientResponse, authGSSClientStep): # pylint: disable=unused-argument,invalid-name
+        """ Check if we can we a Kerberos connection, mocking up the kerberos calls """
+
+        # Setting up the return value
+        authGSSClientResponse.return_value = "Mocking the response"
```
dblock, line 17: Nit: make `return_value` similar to what's typically returned, e.g. `"Negotiate token"`.

dblock (changes requested):
I have some more. I know it's just a small piece of code, but let's get it right! Thanks for hanging in here with me.

psaiz:
Thanks @dblock ! I'm happy to do as many changes as are needed, so please, keep the comments coming. 

I left a reply in the three open discussions. Please, let me know if those are enough to resolve them.

dblock:
> Thanks @dblock ! I'm happy to do as many changes as are needed, so please, keep the comments coming.
> 
> I left a reply in the three open discussions. Please, let me know if those are enough to resolve them.

Thanks. See my two comments above, the unused variable is definitely a bug.

aiudirog:
@psaiz Have you considered just leveraging a standard Kerberos authentication library for requests like [`requests-gssapi`](https://pypi.org/project/requests-gssapi/) or [`requests-kerberos`](https://pypi.org/project/requests-kerberos/)? If so, is there a particular reason you don't want to use them?

If you haven't, you should be able to use them this way:

```python
from opensearchpy import OpenSearch, RequestsHttpConnection
from requests_gssapi import HTTPSPNEGOAuth

opensearch = OpenSearch(
    ...,
    connection_class=RequestsHttpConnection,
    http_auth=HTTPSPNEGOAuth(),
)
```

This is similar to the standard mechanism for [authenticating with AWS IAM roles](https://docs.aws.amazon.com/opensearch-service/latest/developerguide/request-signing.html#request-signing-python).

Making dedicated `Connection` implementations for every combination of HTTP client and authentication mechanism is an N^2 problem that could get out of hand pretty quickly, especially when most clients already have a standard authentication API.

Some of my personal projects are a [port of requests-gssapi to HTTPX](https://github.com/pythongssapi/httpx-gssapi) and adding Windows support to the Python [`gssapi`](https://pypi.org/project/gssapi/1.8.1/) client, so I know my way around this scene a bit.

psaiz:
Hi @aiudirog, thanks for the input. If that approach works, it is indeed much better than creating a new class. Let me play with it and see if it works in my case.

dblock:
Either way, I'd appreciate contributions to https://github.com/opensearch-project/opensearch-py/blob/main/GETTING_STARTED.md for this, please! We might also want to rename that to USER_GUIDE to match other clients, cc: @wbeckler - is there a campaign to standardize those things?

aiudirog:
Can you give the `requests_kerberos` library a try? That one uses pykerberos directly and I'm curious if the issue is GSSAPI or requests' handling of the auth.

psaiz:
Hi @aiudirog ,

The `requests_kerberos` worked, thanks! In my case, something like:
```
from opensearchpy import OpenSearch, RequestsHttpConnection
import requests_kerberos
url = 'htps://...'
certpath = '/etc/...'
client = OpenSearch([url],
                   use_ssl=True,
                   verify_certs=True,
                   ca_certs=certpath,
                   connection_class=RequestsHttpConnection,
http_auth=requests_kerberos.HTTPKerberosAuth(mutual_authentication=requests_kerberos.OPTIONAL)
```
did the trick. And, once I knew the parameters for the mutual_authentication, I went back to `requests_gssapi`, and 
```
...
from requests_gssapi import HTTPSPNEGOAuth, OPTIONAL
...
client = OpenSearch([url],
 ...
http_auth=HTTPSPNEGOAuth(mutual_authentication=OPTIONAL),
)
```
also worked. 

So, indeed, the new class makes no sense. Following the comment from @dblock, I'll change this MR to update the documentation. I was thinking about leaving in a test (mocking the calls), and the optional dependency for the python_requests. Please, let me know if you want me to remove those as well.

USER_GUIDE.md
```diff
@@ -421,4 +427,26 @@ response = client.search(
 
 print('\nSearch results:')
 print(response)
-```
\ No newline at end of file
+```
+
+### Using Kerberos
+
+There are several python packages that provide kerberos support over http connections, like [requests-kerberos](http://pypi.org/project/requests-kerberos) and [requests-gssapi](https://pypi.org/project/requests-gssapi). The following example shows how to setup the authentitcation. Note that some of the parameters, like the `mutual_authentication` might depend on the server settings.
```
dblock, line 434: capitalize Kerberos

dblock: change `like the` to `such as mutual_...`

USER_GUIDE.md
```diff
@@ -421,4 +427,26 @@ response = client.search(
...
+
+```python
+
+from opensearchpy import OpenSearch, RequestsHttpConnection
+from requests_kerberos import HTTPKerberosAuth, OPTIONAL
+
+url = 'htps://...'
+client = OpenSearch(
+    [url],
```
dblock, line 443: nit: since URL is not reused, just put it into the parameters

```python
client = OpenSearch(
    ['https://...'],
```

USER_GUIDE.md
```diff
@@ -421,4 +427,26 @@ response = client.search(
...
+    use_ssl=True,
+    verify_certs=True,
+    connection_class=RequestsHttpConnection,
+    http_auth=HTTPKerberosAuth(mutual_authentication=OPTIONAL)
+)
+
+health = client.cluster.health()
+print(health)
```
dblock, line 451: I would remove `print()` ;)

dblock (changes requested):
The docs are great, and I would be down with just merging that. 

You correctly point out that the test doesn't quite test that the kerberos auth instance is actually called. What else does `send` call? Is there a way to mock those other things? If it's too difficult, then I think we should just remove it. We don't need to test other people's code really, and I believe we already have tests for `http_auth` being used (or we should have some).

psaiz:
Thanks @dblock, changed applied, and test removed

USER_GUIDE.md
```diff
@@ -371,13 +373,17 @@ query = {
 response = client.plugins.alerting.acknowledge_alert(query)
 print(response)
 ```
-## Using IAM credentials for authentication
+## Using different authentication schemas
```
dblock, line 376: It's "scheme" vs. "schema", but I think you mean "methods"

Btw, maybe we can align the Capitalization of Titles in this PR in the guide? We say "Using Kerberos", but also "Using IAM credentials".

USER_GUIDE.md
```diff
@@ -371,13 +373,17 @@ query = {
...
+
+It is possible to use different schemas for the authentication to OpenSearch. The parameters of `connection_class` and `http_auth` can be used for this. As an example, here we have how to authenticate using IAM credentials and Kerberos authentication.
```
dblock, line 378: schemas -> methods

As an example, here we have -> The following example shows how to authenticate using IAM credentials via Kerberos.

USER_GUIDE.md
```diff
@@ -421,4 +427,24 @@ response = client.search(
 
 print('\nSearch results:')
 print(response)
-```
\ No newline at end of file
+```
+
+### Using Kerberos
+
+There are several python packages that provide Kerberos support over http connections, like [requests-kerberos](http://pypi.org/project/requests-kerberos) and [requests-gssapi](https://pypi.org/project/requests-gssapi). The following example shows how to setup the authentication. Note that some of the parameters, such as `mutual_authentication` might depend on the server settings.
```
dblock, line 434: http -> HTTP

like -> such as

dblock (changes requested):
The only must have is schemas -> schemes (or methods). Schemas means something else.

Thanks for hanging in here with me! Really appreciate your help.

psaiz:
Thanks @dblock, ready for the next round

dblock (approved):

USER_GUIDE.md
```diff
@@ -1,4 +1,4 @@
-- [Getting Started with the OpenSearch Python Client](#getting-started-with-the-opensearch-python-client)
+- [Getting started with the OpenSearch python client](#getting-started-with-the-opensearch-python-client)
```
dblock, line 1: I think Python is spelled Capital, but I'll merge as is. 

My biggest gripe with this file is those `** ***` titles ;)

dblock (dismissed):

psaiz:
Ok, sorry that I closed it. I've merged it with the changes that had happened in the mean time, and added a line to the Changelog, since one of the workflows was failing

dblock (approved):

dblock:
@harshavamsi @VachaShah 2nd pls? Let's merge this?

harshavamsi (approved):

PR #226, opened 2022-10-26
Include parsed error info in TransportError in async connections

### LINKED ISSUE #225: [BUG] additional info missing from TransportError when using async
### What is the bug?
`TransportError`s are missing the parsed json stored on the `info` property when using the library with async.

The cause is the content type is not being passed to `_raise_errors` from the async code.
See the difference between the [sync code](https://github.com/opensearch-project/opensearch-py/blob/8879644ec404b3ab15b739b97775d7bd7e51018b/opensearchpy/connection/http_requests.py#L204-L208) and the [async code](https://github.com/opensearch-project/opensearch-py/blob/8879644ec404b3ab15b739b97775d7bd7e51018b/opensearchpy/_async/http_aiohttp.py#L340)

Make bad requests (e.g. 400s) when using the async code.

`TransportError`s `info` property should contain a dict with the parsed response (as in non-async use of the library). 

Linux + MacOs

Links to code above.

This bug was introduced in 1.1.0 when the content-type header started to be inspected before parsing the json. https://github.com/opensearch-project/opensearch-py/commit/1cd15b08453a860624b5c405eed4d4052e916096

gg-kialo:
Fixes #225

Signed-off-by: Gordon Govan <gg@kialo.com>

### Description
Parses the error into the `info` field of `TransportException`s in the async connection to match the behaviour of the sync connection.

### Issues Resolved
Closes #225 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Fixed
 - Fixed import cycle when importing async helpers ([#311](https://github.com/opensearch-project/opensearch-py/pull/311))
 - Fixed userguide for async client ([#340](https://github.com/opensearch-project/opensearch-py/pull/340))
+- Include parsed error info in TransportError in async connections (fixes #225) ([#226](https://github.com/opensearch-project/opensearch-py/pull/226)
 ### Security
 - Fixed CVE-2022-23491 reported in opensearch-dsl-py ([#295](https://github.com/opensearch-project/opensearch-py/pull/295))
 - Update ci workflows ([#318](https://github.com/opensearch-project/opensearch-py/pull/318))
--- a/opensearchpy/_async/http_aiohttp.py
+++ b/opensearchpy/_async/http_aiohttp.py
@@ -331,7 +331,11 @@ class AIOHttpConnection(AsyncConnection):
                 status_code=response.status,
                 response=raw_data,
             )
-            self._raise_error(response.status, raw_data)
+            self._raise_error(
+                response.status,
+                raw_data,
+                response.headers.get("content-type"),
+            )
 
         self.log_request_success(
             method, str(url), url_path, orig_body, response.status, raw_data, duration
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -12,8 +12,9 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Removed
 
 ### Fixed
+- Include parsed error info in TransportError in async connections (TODO)
```
dblock, line 15: Add a link here to the PR 

```
([#226](https://github.com/opensearch-project/opensearch-py/pull/226))
```

dblock (changes requested):
How does this include the error info? I see you add content-type.

Either way needs a test, please.

gg-kialo:
Any hints as to where I could add a test? https://github.com/opensearch-project/opensearch-py/blob/main/test_opensearchpy/test_async/test_connection.py would seem like a good choice, but those tests are using httpbin.org, which doesn't provide an endpoint that returns both a error code and a json body.

dblock:
> Any hints as to where I could add a test? https://github.com/opensearch-project/opensearch-py/blob/main/test_opensearchpy/test_async/test_connection.py would seem like a good choice, but those tests are using httpbin.org, which doesn't provide an endpoint that returns both a error code and a json body.

I generally have a very dumb approach to this, so a change in `opensearchpy/_async/http_aiohttp.py` should be tested in `.../_async/test_http_aiohttp.py`.

dblock (approved):

dblock (changes requested):
One more. I see places (e.g. https://github.com/opensearch-project/opensearch-py/blob/2abb8c43171c2d440d76d26b42972a4bd72996c2/opensearchpy/transport.py#L420) where we do lowercase content-type. Is the map not case-sensitive? Is it always `Content-Type`?

gg-kialo:
@dblock, yes, `headers` is a `CIMultiDictProxy` (where CI is case-insensitive).

dblock (approved):

PR #245, opened 2022-11-23
Modified index definition in client/__init__py

### LINKED ISSUE #132: [BUG] OpenSearch.index() erroneously invokes PUT method resulting in data loss
**What is the bug?**
At https://github.com/opensearch-project/opensearch-py/blob/65c12d7224c9b57d60397f1069e4cbaa1caea6d6/opensearchpy/client/__init__.py#L358

We do: `"POST" if id in SKIP_IN_PATH else "PUT"`

Logically this means "If no id, POST" therefore "If id, PUT"

According to the API reference here: https://opensearch.org/docs/latest/opensearch/index-data/

> POST requests make partial updates to documents. To altogether replace a document, use a PUT request

If id is provided, and the user intends to update a document, but PUT is used, the document will be overwritten. The expression should be the inverse. Anecdotally we observed this in our implementation.

1. Index a document

```
# instantiate OpenSearch as client
# index a document
client.index(
    index=movies,
    id=1234,
    body={
      'doc': {
        'title': "Castle in the Sky",
        'genre': ["Animation", "Fantasy"]
      }
    }
)
```

2. Update the document

```
# update document
client.index(
    index=movies,
    id=1234,
    body={
      'doc': {
        'genre': ["Animation", "Fantasy"]
      }
    }
)
# the document lacks the title property, having been overwritten
```

When an id is provided, the POST method is used to update a document in accordance with the API reference. A partial update does not overwrite the entire document.

 - OS: macOS 
 - Version 10.15.7 (19H1615)
 - Plugins

not applicable

In actuality the REST API does not seem to conform to the behavior described in the documentation. Whether POST or PUT is used, the document is overwritten. That is perhaps why this bug hasn't been raised before: people just use the `OpenSearch.update()` method. One solution might be to fix the documentation and maybe remove the erroneous line, because in effect it does nothing. To bring everything in line with the expected and currently documented behavior, both the API and this client (and other clients?) need to be fixed.

saimedhi:

### Description
Modified index definition in client/__init__py

### Issues Resolved
Closes #132 

### DIFF
--- a/opensearchpy/client/__init__.py
+++ b/opensearchpy/client/__init__.py
@@ -312,7 +312,7 @@ class OpenSearch(object):
     )
     def index(self, index, body, id=None, params=None, headers=None):
         """
-        Creates or updates a document in an index.
+        Creates or overwrites a document in an index.
 
 
         :arg index: The name of the index


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Changed
 - Updated getting started to user guide ([#233](https://github.com/opensearch-project/opensearch-py/pull/233))
 - Updated CA certificate handling to check OpenSSL environment variables before defaulting to certifi ([#196](https://github.com/opensearch-project/opensearch-py/pull/196))
+- Modified index definition in client/__init__py([#245](https://github.com/opensearch-project/opensearch-py/pull/245))
```
dblock, line 12: No need for a CHANGELOG entry for a doc change, you can remove. It needs an extra space before `(` otherwise I would have merged anyway :)

dblock (changes requested):
I'm being nitpicky. Forgive me?

dblock (approved):

PR #254, opened 2022-12-19
Adding async support for AWSSigV4

### LINKED ISSUE #131: [FEATURE] Async support
Is the async feature stable to use? I tried following code but get error "AttributeError: 'AWS4Auth' object has no attribute 'encode'"

```
async_search = AsyncOpenSearch(
    hosts = [{'host': host, 'port': 443}],
    http_auth = awsauth,
    use_ssl = True,
    verify_certs = True,
    transport_class=AsyncTransport,
)

async_search.search(sth...)
```

harshavamsi:
Signed-off-by: Harsha Vamsi Kalluri <harshavamsi096@gmail.com>

### Description
Adds support for Async Sigv4 to authenticate with AWS IAM.

### Issues Resolved
Closes #131 

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -1,6 +1,6 @@
-- [Getting Started with the OpenSearch Python Client](#getting-started-with-the-opensearch-python-client)
+- [User guide of OpenSearch Python Client](#user-guide-of-opensearch-python-client)
   - [Setup](#setup)
-  - [Sample code](#sample-code)
+  - [Example](#example)
     - [Creating a client](#creating-a-client)
     - [Creating an index](#creating-an-index)
     - [Adding a document to an index](#adding-a-document-to-an-index)
@@ -10,7 +10,7 @@
     - [Deleting a document](#deleting-a-document)
     - [Deleting an index](#deleting-an-index)
   - [Making API Calls](#making-api-calls)
-    - [Point in Time API](#point-in-time-api-calls)
+    - [Point in Time API](#point-in-time-api)
   - [Using plugins](#using-plugins)
     - [Alerting plugin](#alerting-plugin)
       - [**Searching for monitors**](#searching-for-monitors)
@@ -21,6 +21,7 @@
       - [**Acknowledge alerts**](#acknowledge-alerts)
   - [Using IAM credentials for authentication](#using-iam-credentials-for-authentication)
       - [Pre-requisites to use `AWSV4SignerAuth`](#pre-requisites-to-use-awsv4signerauth)
+  - [Using IAM authentication with Async opensearch-py](#using-iam-authentication-with-async-opensearch-py)
 
 # User guide of OpenSearch Python Client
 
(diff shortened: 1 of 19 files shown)


USER_GUIDE.md
```diff
@@ -439,6 +440,51 @@ print('\nSearch results:')
 print(response)
 ```
 
+## Using IAM authentication with Async opensearch-py
+
+Make sure to use the new `RequestsAsyncHttpConnection` connection class with the new async `AWSV4SignerAuthAsync` signer.
```
dblock, line 445: Nit: Since it's a user guide I would remove "new". So just "Use `Request...` with the async `AWSV...` signer."

opensearchpy/__init__.py
```diff
@@ -95,6 +97,7 @@
     "OpenSearchWarning",
     "OpenSearchDeprecationWarning",
     "AWSV4SignerAuth",
+    "AWSV4SignerAuthAsync",
```
dblock, line 100: Should it be `AWSV4SignerAsyncAuth` to match naming in `RequestsAsyncHttpConnection`?

opensearchpy/connection/http_async_requests.py
```diff
@@ -0,0 +1,287 @@
...
+        url_path = self.url_prefix + url
+        if params:
+            query_string = urlencode(params)
+        else:
+            query_string = ""
+
+
+        # There is a bug in aiohttp that disables the re-use
+        # of the connection in the pool when method=HEAD.
+        # See: aio-libs/aiohttp#1769
```
dblock, line 161: Nit: Make it a full link to make it clickable.

opensearchpy/connection/http_async_requests.py
```diff
@@ -0,0 +1,287 @@
...
+                method,
+                url,
+                data=body,
+                headers=req_headers,
+                timeout=timeout,
+                fingerprint=self.ssl_assert_fingerprint,
+            ) as response:
+                if is_head:  # We actually called 'GET' so throw away the data.
+                    await response.release()
+                    raw_data = ""
```
dblock, line 208: Should it be `None`?

dblock (changes requested):
I have some nits/asks.

Consider expanding tests. For example `asyncsigner.py` should have its own `test_asyncsigner.py` for all kind of really basic functions to avoid regressions and have a test set to build on top of in the future.

opensearchpy/connection/http_async_requests.py
harshavamsi: https://github.com/harshavamsi/opensearch-py/blob/0ec369ffd5632191d1f4b1c6191bb4d381f2f9c2/opensearchpy/_async/http_aiohttp.py#L290 We're inheriting the upstream class, looks like it's instantiated the same way there.

USER_GUIDE.md
```diff
@@ -439,6 +440,51 @@ print('\nSearch results:')
 print(response)
 ```
 
+## Using IAM authentication with Async opensearch-py
```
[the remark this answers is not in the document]
harshavamsi: Fixed this.

USER_GUIDE.md
harshavamsi: Makes sense. Fixed.

opensearchpy/__init__.py
harshavamsi: Yep, I updated the naming on that. I updated the connection class to `AsyncHttpConnection` as well because I realized it was actually not using the requests library.

opensearchpy/connection/http_async_requests.py
harshavamsi: Updated to full link.

harshavamsi:
> I have some nits/asks.
> 
> Consider expanding tests. For example `asyncsigner.py` should have its own `test_asyncsigner.py` for all kind of really basic functions to avoid regressions and have a test set to build on top of in the future.

@dblock this PR is ready for review now.  I created a new test class for the async signer. We can add more tests to that namespace in the future. Tested this locally with an async client and it works as expected.

dblock (approved):

dblock:
I like it. If you have time, you could make an async version of https://github.com/dblock/opensearch-python-client-demo (Pipfile lets you use a relative path with `path` for the dependency).

PR #257, opened 2022-12-31
Added new OpenSearch versions and updated compatibility matrix

### LINKED ISSUE #252: [BUG] Client/Core Engine Compatibility Matrix Wrong/Outdated
### What is the bug?
* The compatibility matrix [1] indicates that I need client v1.1 to talk to OS 1.2.0, but client v1.0 to talk to OS 1.2.1.  This is confusing and almost certainly not intended.
* Additionally, the compatibility matrix gives no indication on compatibility with the many newer versions of the core engine (2.1, 2.2, 2.3, 2.4).  It would be helpful to indicate whether the newest version of the client (2.0) should be considered compatible or not, or at least whether that's currently unknown/untested.  Right now, readers are left with no idea whatsoever, especially since the matrix is almost certainly wrong about compatibility with OS 1.X versions.

[1] https://github.com/opensearch-project/opensearch-py/blob/main/COMPATIBILITY.md

Read the compatibility file :-)

N/A

N/A

N/A

N/A

saimedhi:

### Description
Added new OpenSearch versions and updated compatibility matrix

### Issues Resolved
Closes #252 


wbeckler:
Do you know if client v2.0 is compatible with older versions of the server, such as 1.3?

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added overload decorators to helpers-actions.pyi-"bulk" ([#239](https://github.com/opensearch-project/opensearch-py/pull/239))
 - Document Keberos authenticaion ([214](https://github.com/opensearch-project/opensearch-py/pull/214))
 - Add release workflows ([#240](https://github.com/opensearch-project/opensearch-py/pull/240))
+- Added new OpenSearch versions and updated compatibility matrix ([#257](https://github.com/opensearch-project/opensearch-py/pull/257))
```
dblock, line 12: We should be explicit here and say something like "Compatibility with OpenSearch 2.1.0 - 2.4.1".

dblock (dismissed):
I'm ok without a better changelog, but I think users would appreciate it.

dblock:
> Do you know if client v2.0 is compatible with older versions of the server, such as 1.3?

It should be according to https://github.com/opensearch-project/opensearch-py/blob/main/.github/workflows/integration.yml#L14.

wbeckler:
@saimedhi I would suggest adding in the older versions  to the matrix that are compatible, as the presumption for users would be that the lack of a version means there exists an incompatibility.

harshavamsi (dismissed):
LGTM!

COMPATIBILITY.md
```diff
@@ -5,23 +5,15 @@
...
-| 1.3.1 | 1.1.0 |
-| 1.3.2 | 1.1.0 |
-| 1.3.3 | 1.1.0 |
-| 2.0.0 | 2.0.0 |
-| 2.0.1 | 2.0.0 |
+Compatibility 
+
+| Client Version | OpenSearch Version | Notes |
+| --- | --- | --- |
+| 1.0.0 | 1.0.0-1.0.1, 1.2.1-1.2.4 | |
```
VachaShah, line 12: It might be a typo in the previous markdown, 1.1.0 client version was released with being compatible from OpenSearch versions 1.3.0. Can you please update that?

saimedhi: @VachaShah Made the changes as suggested.

harshavamsi (approved):

PR #271, opened 2023-01-17
Added support for skip-changelog.

dblock:
Created a skip-changelog label and skip CHANGELOG on it.


dblock:
This was a product of automation, noop here.

PR #280, opened 2023-02-07
Merging opensearch-dsl-py to opensearch-py

### LINKED ISSUE #194: [PROPOSAL] Merge opensearch-dsl-py into this repository
### What are you proposing?

Can't think of any. 

saimedhi:

### Description
Merging opensearch-dsl-py to opensearch-py

### Issues Resolved
Closes #194 


saimedhi:
Please go through the issue mentioned above for additional context.

dblock:
I think this looks good. What do you think about creating an UPGRADING.md like https://github.com/slack-ruby/slack-ruby-client/blob/master/UPGRADING.md?

saimedhi:
> I think this looks good. What do you think about creating an UPGRADING.md like https://github.com/slack-ruby/slack-ruby-client/blob/master/UPGRADING.md?

Ok, will do that. Adding this to the list of things to do. @dblock please take a look at this [list](https://github.com/opensearch-project/opensearch-py/issues/194#issuecomment-1420287048) and let me know if any modifications are required.

opensearchpy/connection/http_async.py
```diff
@@ -7,6 +7,23 @@
 # Modifications Copyright OpenSearch Contributors. See
 # GitHub history for details.
 #
+#  Licensed to Elasticsearch B.V. under one or more contributor
```
harshavamsi, line 10: These licenses headers are not required because this file was created post fork.

opensearchpy/connection/http_async.pyi
```diff
@@ -6,6 +6,23 @@
 #
 # Modifications Copyright OpenSearch Contributors. See
 # GitHub history for details.
+#
```
harshavamsi, line 9: These licenses headers are not required because this file was created post fork.

opensearchpy/exceptions.py
```diff
@@ -163,6 +169,22 @@ class AuthorizationException(TransportError):
     """Exception representing a 403 status code."""
 
 
+class OpenSearchDslException(Exception):
+    """OpenSearchDsl Exception"""
+
+
+class UnknownDslObject(OpenSearchDslException):
+    """OpenSearchDsl Exception"""
```
harshavamsi, line 177: These exception messages need to be updated.

opensearchpy/exceptions.py
```diff
@@ -163,6 +169,22 @@ class AuthorizationException(TransportError):
...
+
+
+class ValidationException(ValueError, OpenSearchDslException):
+    """OpenSearchDsl Exception"""
```
harshavamsi, line 181: Same as above.

opensearchpy/exceptions.py
```diff
@@ -163,6 +169,22 @@ class AuthorizationException(TransportError):
...
+
+
+class IllegalOperation(OpenSearchDslException):
+    """OpenSearchDsl Exception"""
```
harshavamsi, line 185: Same as above.

opensearchpy/transport.py
```diff
@@ -197,13 +197,14 @@ def set_connections(self, hosts):
...
         """
+
         # construct the connections
         def _create_connection(host):
             # if this is not the initial setup look at the existing connection
             # options and identify connections that haven't changed and can be
             # kept around.
             if hasattr(self, "connection_pool"):
-                for (connection, old_host) in self.connection_pool.connection_opts:
+                for connection, old_host in self.connection_pool.connection_opts:
```
harshavamsi, line 207: Why make this change?

opensearchpy/connection/http_async.py
saimedhi: Made the change suggested.

opensearchpy/connection/http_async.pyi
saimedhi: Made the change suggested.

opensearchpy/exceptions.py
saimedhi: Made the change suggested.

opensearchpy/exceptions.py
saimedhi: Made the change suggested.

opensearchpy/exceptions.py
saimedhi: Made the change suggested.

opensearchpy/transport.py
saimedhi: linter is automatically formatting it. Without it lint error is produced.

.gitignore
```diff
@@ -141,4 +141,18 @@ cython_debug/
 # opensearch files
 test_opensearch/cover
 test_opensearch/local.py
-.ci/output
\ No newline at end of file
+.ci/output
+
+# opensearch-dsl .gitignore
```
VachaShah, line 146: Lets merge the gitignore paths, there are some that are repeated in this section.

VachaShah (commented):
How about moving the files into folders like `client`, `helper` etc. which exist inside `opensearchpy` instead of moving them directly under `opensearchpy`?

saimedhi:
> How about moving the files into folders like `client`, `helper` etc. which exist inside `opensearchpy` instead of moving them directly under `opensearchpy`?

@VachaShah I will do it in next PR while creating async version of dsl. But as of now to minimize the differences for existing opensearch-dsl-py users it is left under opensearchpy. Later I will place it as separate folder under opensearchpy and opensearchpy/async.

.gitignore
saimedhi: Removed repeated lines and merged gitignore paths

saimedhi:
Closing this PR because of merge conflicts and raised PR #287.

Issue #281, opened 2023-02-07
[BUG] Cannot import AsyncOpenSearch in Linux Container


cdobbsTM:
### What is the bug?
`ImportError: cannot import name 'AsyncOpenSearch' from 'opensearchpy' (/app/.local/lib/python3.9/site-packages/opensearchpy/__init__.py)` when running within a Linux Docker container

### How can one reproduce the bug?
1. Create Docker image from ubi8 base image
2. Install Python 3.9
3. Install opensearch-py==2.1.1
4. Run "from opensearchpy import AsyncOpenSearch" from within container

### What is the expected behavior?
The AsyncOpenSearch client to be successfully imported

### What is your host/environment?
Windows 10, RedHat ubi8 container

### Do you have any screenshots?
Not necessary 

### Do you have any additional context?
When I run my project outside of Docker it works fine.

Code from `opensearchpy\__init__.py`:
```
try:
    # Asyncio only supported on Python 3.6+
    if sys.version_info < (3, 6):
        raise ImportError

    from ._async.client import AsyncOpenSearch
    from ._async.http_aiohttp import AIOHttpConnection, AsyncConnection
    from ._async.transport import AsyncTransport
    from .connection import AsyncHttpConnection

    __all__ += [
        "AIOHttpConnection",
        "AsyncConnection",
        "AsyncTransport",
        "AsyncOpenSearch",
        "AsyncHttpConnection",
    ]
except (ImportError, SyntaxError):
    pass
```
I imagine this code might have something to do with the issue but I don't see why it would skip the imports when running on Python 3.9.

cdobbsTM:
#96 I can see that this issue mentions the same problem.

Trying `opensearch-py[async]==2.1.1`. 

Although I don't see why there is a change in behaviour between installing into a container vs on the host machine.

cdobbsTM:
The above resolved this but I can't see it documented anywhere.

wbeckler:
Would you be up for submitting a documentation PR?

PR #286, opened 2023-02-13
add async reference to user guide

cdobbsTM:
as per the suggestion by @wbeckler on issue #281

### DIFF
--- a/USER_GUIDE.md
+++ b/USER_GUIDE.md
@@ -33,12 +33,14 @@ To add the client to your project, install it using [pip](https://pip.pypa.io/):
 
 ```bash
 pip install opensearch-py
+pip install opensearch-py[async] # to include async client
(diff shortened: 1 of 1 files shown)


dblock (commented):
You never want both sync and async, you must choose whether you do async vs. non-async code if I understand this correctly. I propose we add a "Creating an Async Client" or some kind of section like that? WDYT?

dblock:
You'll need to fix DCO with `git commit -s`, please, too.

saimedhi:
Hello, @cdobbsTM. Thank you for your contribution to the opensearch-py repo. The following [PR](https://github.com/opensearch-project/opensearch-py/pull/340/files) adds an async reference to the user guide. If you agree to close this PR, please let me know. Thanks:)

PR #287, opened 2023-02-13
Merging opensearch-dsl-py to opensearch-py

### LINKED ISSUE #194: [PROPOSAL] Merge opensearch-dsl-py into this repository

saimedhi:

### Description
Merging opensearch-dsl-py to opensearch-py

### Issues Resolved
Closes #194  


USER_GUIDE.md
```diff
@@ -228,6 +230,33 @@ print('\n The deleted point in time:')
 print(response)
 ```
 
+## Using High-level Python client
```
harshavamsi, line 233: Is this the only piece that is being carried forward from the high level client?

harshavamsi (changes requested):
Can you also update the documentation to include the new DSL helpers in the `docs` folder?

USER_GUIDE.md
saimedhi: All other API's added have similar implementation. So, I added only Search API example.

harshavamsi (approved):

adilnaimi:
When will this PR be released?

VachaShah:
> When will this PR be released?

@harshavamsi Can you comment here?

harshavamsi:
> > When will this PR be released?
> 
> @harshavamsi Can you comment here? Tracking in #307

Released here -- https://github.com/opensearch-project/opensearch-py/releases/tag/v2.2.0

PR #293, opened 2023-02-15
Added upgrading.md file and updated it for opensearch-py 2.2.0 release

### LINKED ISSUE #194: [PROPOSAL] Merge opensearch-dsl-py into this repository

saimedhi:
### Description
Added upgrading.md file and updated it for opensearch-py 2.2.0 release

### Issues Resolved
Closes #194 

### DIFF
--- a/UPGRADING.md
+++ b/UPGRADING.md
@@ -1,0 +1,24 @@
+- [Upgrading OpenSearch Python Client](#upgrading-opensearch-python-client)
+  - [Upgrading to >= 2.2.0](#upgrading-to->=-2.2.0)
+    - [Features from High-Level Python Client](#features-from-high-level-python-client)
+    - [API Implementation differences for existing opensearch-dsl-py users](#api-implementation-differences-for-existing-opensearch-dsl-py-users)
+
+
+# Upgrading OpenSearch Python Client
+
+## Upgrading to >= 2.2.0
+
+opensearch-py now includes [opensearch-dsl-py](https://pypi.org/project/opensearch-dsl/) features. opensearch-dsl-py was merged into opensearch-py preserving backwards compatibility with the previous opensearch-py version. (Refer [link](https://github.com/opensearch-project/opensearch-py/pull/287))
+
+
+### Features from High-Level Python Client
+
+opensearch-py functionalities that already exist are not altered. 
+
+The opensearchpy [helpers](https://github.com/opensearch-project/opensearch-py/tree/main/opensearchpy/helpers) module now provides access to aggs, analysis, document, faceted search, field, function, index, mapping, query, search, update by query, utils, and wrappers. As a result, without importing opensearch-dsl-py, these functionalities can be imported directly from opensearch-py. 
+
+
+### API Implementation differences for existing opensearch-dsl-py users 
+
+The functionalities from opensearch-dsl-py are merged into this client. Refer to the [USER_GUIDE](https://github.com/opensearch-project/opensearch-py/blob/main/USER_GUIDE.md#using-high-level-python-client) for an example on how to implement a feature from the previously known high level Python client.
+
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,12 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [2.1.1]
 ### Added
+- Merging opensearch-dsl-py into opensearch-py ([#287](https://github.com/opensearch-project/opensearch-py/pull/287))
+- Added upgrading.md file and updated it for opensearch-py 2.2.0 release ([#293](https://github.com/opensearch-project/opensearch-py/pull/293))
 ### Changed
 ### Deprecated
 ### Removed
+- Removed 'out/opensearchpy' folder which was produced while generating pyi files for plugins ([#288](https://github.com/opensearch-project/opensearch-py/pull/288))
 ### Fixed
 - Fixed SigV4 Signing for Managed Service ([#279](https://github.com/opensearch-project/opensearch-py/pull/279))
 - Fixed SigV4 Signing for Async Requests with QueryStrings ([#272](https://github.com/opensearch-project/opensearch-py/pull/279))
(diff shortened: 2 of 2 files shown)


upgrading.md
```diff
@@ -0,0 +1,36 @@
+Upgrading OpenSearch Python Client
+==================================
+
+### Upgrading to >= 2.2.0
+
+Opensearch-py now includes opensearch-dsl-py features. Opensearch-dsl-py was merged into opensearch-py in the following PR [#287](https://github.com/opensearch-project/opensearch-py/pull/287).
```
VachaShah, line 6: * Spell as `opensearch-py` and `opensearch-dsl-py`.
* Add the pypi link for opensearch-dsl-py.
* You can also say "opensearch-dsl-py` was merged into `opensearch-py` preserving backwards compatibility with the previous `opensearch-py` version. (Refer link)"

upgrading.md
```diff
@@ -0,0 +1,36 @@
+Upgrading OpenSearch Python Client
```
VachaShah, line 1: Add a TOC

upgrading.md
```diff
@@ -0,0 +1,36 @@
...
+### Upgrading to >= 2.2.0
+
+Opensearch-py now includes opensearch-dsl-py features. Opensearch-dsl-py was merged into opensearch-py in the following PR [#287](https://github.com/opensearch-project/opensearch-py/pull/287). 
+
+
+#### Overview of architectural changes:
+
+Opensearch-py functionalities that already exist are not altered. Only the following areas have new additions.
+
+The opensearchpy "helpers" module now provides access to aggs, analysis, document, faceted search, field, function, index, mapping, query, search, update by query, utils, and wrappers. As a result, without importing opensearch-dsl-py, these functionalities can be imported directly from opensearch-py. The "connections" file from DSL is now included in the "connection" module. Additionally, the dsl files "serializer" and "exceptions" are combined with their equivalent opensearchpy files "serializer" and "exceptions".
```
VachaShah, line 13: * Add a link to the helpers module.
* Add a link to the user guide where users can see how to use the new APIs.
* I think its okay if we don't mention how the files were merged.

upgrading.md
```diff
@@ -0,0 +1,36 @@
...
+
+
+#### API Implementation differences for existing opensearch-dsl-py users 
+
+As mentioned above, Opensearch-py functionalities that already exist are not altered. Switching from opensearch-dsl-py to opensearch-py is made simple by minimising API implementation differences. Refer to the [userguide](https://github.com/opensearch-project/opensearch-py/blob/main/USER_GUIDE.md#using-high-level-python-client) for an example of how to implement a newly added feature.
```
VachaShah, line 18: We can say "The functionalities from opensearch-dsl-py are merged into this client. Refer to the [USER_GUIDE](https://github.com/opensearch-project/opensearch-py/blob/main/USER_GUIDE.md#using-high-level-python-client) for an example on how to implement a feature from the previously known high level Python client."

upgrading.md
```diff
@@ -0,0 +1,36 @@
...
+Before:
+```python
+from opensearchpy import OpenSearch
+    from opensearch_dsl import Search
+```
+
+Now:
+```python
+from opensearchpy import OpenSearch, Search
+```
```
VachaShah, line 31: If this is already mentioned in the user guide, it might be redundant to add it here.

upgrading.md
```diff
@@ -0,0 +1,36 @@
...
+
+
+#### Deprecated Methods
+
+No opensearch-py methods are deprecated with this release. However, the opensearch-dsl-py library will be deprecated.
```
VachaShah, line 36: I think we can remove this section to avoid confusion. The deprecation of dsl-py can be noted in its repo.

upgrading.md
```diff
@@ -0,0 +1,36 @@
+Upgrading OpenSearch Python Client
+==================================
+
+### Upgrading to >= 2.2.0
+
+Opensearch-py now includes opensearch-dsl-py features. Opensearch-dsl-py was merged into opensearch-py in the following PR [#287](https://github.com/opensearch-project/opensearch-py/pull/287). 
+
+
+#### Overview of architectural changes:
```
VachaShah, line 9: Can we call this something like "Features from High-Level Python Client"?

upgrading.md
dblock, line 1: Let's call the file UPGRADING.md all caps like other markdowns.

UPGRADING.md
```diff
@@ -0,0 +1,24 @@
+- [Upgrading OpenSearch Python Client](#upgrading-opensearch-python-client)
+  - [Upgrading to >= 2.2.0](#upgrading-to->=-2.2.0)
```
harshavamsi, line 2: This link needs to be fixed here.

UPGRADING.md
```diff
@@ -0,0 +1,24 @@
...
+# Upgrading OpenSearch Python Client
+
+## Upgrading to >= 2.2.0
+
+opensearch-py now includes [opensearch-dsl-py](https://pypi.org/project/opensearch-dsl/) features. opensearch-dsl-py was merged into opensearch-py preserving backwards compatibility with the previous opensearch-py version. (Refer [link](https://github.com/opensearch-project/opensearch-py/pull/287))
+
+
+### Features from High-Level Python Client
+
+opensearch-py functionalities that already exist are not altered. Only the following areas have new feature additions.
```
harshavamsi, line 16: Can you make this some sort of a bulleted list? Or something that shows "following areas" as areas?

harshavamsi (changes requested):

UPGRADING.md
saimedhi: Vamsi, I think that link is working. I checked in VS code it is pointing rightly. But when we see it in github page it is appearing like it is not working.

UPGRADING.md
saimedhi: Removed following areas line. Thanks for correcting.

harshavamsi (approved):

Issue #313, opened 2023-03-07
[BUG] scan throws KeyError: 'hits' when no results are found


samodelkinas:
https://github.com/opensearch-project/opensearch-py/blob/05326ba5474c1a1f5c1975760a4e37037f967dda/opensearchpy/helpers/actions.py#L570 throws an exception when scroll is used and search results are empty (hits key is not within hits dictionary). 
For example when used with AWS wrangler opensearch.search method:
```
df = wr.opensearch.search(
        client=client,
        index='storage_v3',
        search_body={
            "query": {
                "term": {
                    "feed_id": {
                        "value": "20230307"
                    }
                    
                }
            }
        },
        is_scroll=True)
```
will cause exception:
  File "/opt/airflow/venv/wrangler/lib64/python3.8/site-packages/awswrangler/opensearch/_read.py", line 120, in search
    documents = [_hit_to_row(doc) for doc in documents_generator]
  File "/opt/airflow/venv/wrangler/lib64/python3.8/site-packages/awswrangler/opensearch/_read.py", line 120, in <listcomp>
    documents = [_hit_to_row(doc) for doc in documents_generator]
  File "/opt/airflow/venv/wrangler/lib64/python3.8/site-packages/opensearchpy/helpers/actions.py", line 570, in scan
    while scroll_id and resp["hits"]["hits"]:
KeyError: 'hits'
This can be handled, though would be nice to have empty results instead so that upstream library returns empty dataframe or something.

ReinGrad:
That there is a problem with the opensearchpy library actions.py file. In particular, in line 570 there is a while loop that checks for scroll_id and non-empty search results in the hits key of the corresponding dictionary. However, when using the library with the AWS Wrangler opensearch.search method and the search query does not return results, the hits key is missing from the resp dictionary, which leads to a KeyError exception.

To reproduce the problem, one would have to use the AWS Wrangler opensearch.search method to execute a search query that returns no results, and also use the is_scroll=True argument to enable the scroll function. This would cause a while loop on line 570 actions.py and results in a KeyError exception.

A possible behavior in this case would be if the opensearchpy library handled the absence of the hits key in the resp dictionary and returned empty search results, instead of throwing a KeyError exception.

### Decisions

[...]

PR #314, opened 2023-03-07
Handle empty response

samodelkinas:
https://github.com/opensearch-project/opensearch-py/issues/313

### Description
Handles scroll with no search results (no hits in response) to prevent exception.

### Issues Resolved
resolves https://github.com/opensearch-project/opensearch-py/issues/313

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,9 +3,12 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [2.1.1]
 ### Added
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -3,9 +3,12 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [2.1.1]
```
dblock, line 4: Should this say "Unreleased"?

dblock (changes requested):
I think you pushed different code to this `patch-1` branch than what you intended to :) I do think we want to fix up CHANGELOG and UPGRADING!

PR #321, opened 2023-03-14
Async support for helpers that are merged from opensearch-dsl-py

### LINKED ISSUE #194: [PROPOSAL] Merge opensearch-dsl-py into this repository

### LINKED ISSUE #320: [FEATURE] - async support
**Is your feature request related to a problem?**
Guys, do you have plan to introduce async support in opensearch-dsl (opensearch-py is already doing this) ?

saimedhi:
### Description
Async support for helpers that are merged from opensearch-dsl-py

### Issues Resolved
Closes #320 
Closes #194 


harshavamsi:
@saimedhi can you resolve conflicts?

PR #345, opened 2023-03-29
#283 Replaced line 192-196 with given code from github

### LINKED ISSUE #283: [BUG] initializing an async client without http_auth results in an error
### What is the bug?
With local/tests setup of the library, without any http auth, the client produces the following error on request:
`File ".../.venv/lib/python3.9/site-packages/opensearchpy/connection/http_async.py", line 194, in perform_request
    **self._http_auth(method, url, query_string, body),
TypeError: 'NoneType' object is not callable`

Initialize and use AsyncOpenSearch client without http_auth (i.e with use of local ES on docker for example) and make a call to `async_bulk`

Ignore http_auth if it is None

MacOS 12.5.1
Python 3.9.5
opensearch-py 2.1.1

The fix should be relatively easy
on `opensearchpy/connection/http_async.py` line 194 we should add http_auth call to the request headers only if it is not None
i.e replace lines 192-195 with

```
    req_headers.update(**self._http_auth(method, url, query_string, body))
```

mron03:
### I have looked at issue #283, and as I understood, did what the issue asked, but If I am wrong please correct me, 
so basically I just replaced some code with the given one from the issue

### Issues Resolved
#283 Issue

### DIFF
--- a/opensearchpy/connection/http_async.py
+++ b/opensearchpy/connection/http_async.py
@@ -190,10 +190,8 @@ class AsyncHttpConnection(AIOHttpConnection):
             body = self._gzip_compress(body)
             req_headers["content-encoding"] = "gzip"
 
-        req_headers = {
-            **req_headers,
-            **self._http_auth(method, url, query_string, body),
-        }
+        if self._http_auth is not None:
+            req_headers.update(**self._http_auth(method, url, query_string, body))
 
         start = self.loop.time()
         try:


mron03:
If i did not understand the task right  from the issue, please tell me, i will get everything right

dblock:
This needs a test and a CHANGELOG entry, please?

Issue #347, opened 2023-03-31
[BUG] Tests failing on Windows


sekerbek:
### What is the bug?
Seems there are bugs, while running the test suite `nox -rs test` on a Windows machine.
It gives an failed error - **nox > * test-3.11: failed.**

To solve it, I tried to add appropriate python version in noxfile.py
`@nox.session(python=["2.7", "3.4", "3.5", "3.6", "3.7", "3.8", "3.9", "3.10", "3.11"])`

### What is the expected behavior?
The output should indicate that the tests have completed, showing the number of tests run and how long it took to run them. The final line should indicate that all the tests passed, with the "OK" message.

### What is your host/environment?
Python 3.11, Microsoft Windows 10

### Do you have any screenshots?
![image](https://user-images.githubusercontent.com/79473033/229105422-2da28dbb-d86d-40c1-82ad-8ff87888ca2a.png)
![image](https://user-images.githubusercontent.com/79473033/229105439-150c00fc-a6a9-4bec-8e1d-9e6f715156ee.png)

harshavamsi:
Thanks for bringing this up, I believe @saimedhi has been able to replicate it and is working on it.

saimedhi:
I reproduced the issue on Windows. The problem is not limited to Python 3.11; it is present in all versions beginning with 3.6. (Async support begins with version 3.6). Some async tests are failing. I did not delve further into why they are failing. If anyone is interested in working on this issue, please do so and also mention that you are working on it in the comment. If no one is working on it, I will begin working on it.

dblock:
First step is to add a failing Windows CI into this project, if anyone wants to start with that, it'd be helpful.

margulanz:
I will work on this issue

dblock:
I ran into this trying to add Windows to CI, opened #391 (will close as dup).

dblock:
```
2023-05-15T20:39:04.2655572Z test_opensearchpy/test_transport.py::TestTransport::test_failed_connection_will_be_marked_as_dead FAILED [ 17%]

2023-05-15T20:39:04.3356121Z test_opensearchpy/test_transport.py::TestTransport::test_resurrected_connection_will_be_marked_as_live_on_success FAILED [ 18%]

2023-05-15T20:39:05.7608976Z test_opensearchpy/test_async/test_connection.py::TestAIOHttpConnection::test_uses_given_ca_certs FAILED [ 22%]
[...]

PR #371, opened 2023-04-18
Test

margulanz:
### Description
Updated connection_pool.py, so TypeError does not occur. Updated tests accordingly. I am not sure if this is the best solution, so I am waiting for the advices if there is something wrong with my solution

### Issues Resolved
Partially Closes [347](https://github.com/opensearch-project/opensearch-py/issues/347). 

### DIFF
--- a/opensearchpy/connection_pool.py
+++ b/opensearchpy/connection_pool.py
@@ -174,6 +174,7 @@ class ConnectionPool(object):
         """
         # allow inject for testing purposes
         now = now if now else time.time()
+        print(id(connection), now)
         try:
             self.connections.remove(connection)
         except ValueError:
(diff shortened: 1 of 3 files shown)


opensearchpy/connection_pool.py
```diff
@@ -187,7 +187,7 @@ def mark_dead(self, connection, now=None):
             dead_count = self.dead_count.get(connection, 0) + 1
             self.dead_count[connection] = dead_count
             timeout = self.dead_timeout * 2 ** min(dead_count - 1, self.timeout_cutoff)
-            self.dead.put((now + timeout, connection))
+            self.dead.put((now + timeout,id(connection),connection))
```
dblock, line 190: add a space before `,`

dblock (changes requested):
This changes what we put into the connection pool `dead` set, which in a way changes an interface. Probably not the best solution, so let's think about it a bit more. What's the actual problem? Are you able to explain why the test fails only on Windows?

Add Windows CI into this PR?

saimedhi:
Hey @margulanz, Could you kindly answer the questions mentioned above? Thanks :)

saimedhi:
@margulanz, Could you please respond to the above comments when you have a moment? Thank you!

dblock:
This seems to have been a red herring, fixed windows CI in https://github.com/opensearch-project/opensearch-py/pull/569.

PR #376, opened 2023-04-25
Enhanced existing API generator to use the OpenAPI spec as the input format

### LINKED ISSUE #348: [FEATURE] Enhance the existing code generator to use the OpenAPI spec as the input format
As part of an existing campaign to generate clients off of specs, we'd like to generate new APIs in the python client from an OpenAPI spec. The current generator makes use of the legacy rest-api-spec that is published in core opensearch. There are other initiatives that will convert the rest-api-spec into OpenAPI specs through a translation layer, but for the purposes of this client we'd like to do a re-write on the existing generator to inject OpenAPI.

A rewrite of the current [generator](https://github.com/opensearch-project/opensearch-py/blob/main/utils/generate-api.py) to ingest OpenAPI specs and output API endpoints into the client.  

Some alternatives could be to make use of the existing rest-api-spec and continue using the generator as is. But this is out of the scope of consideration as the spec was written to suit a particular use-case that is no longer relevant. OpenAPI specs are the industry standard for JSON based schemas. We've already seen an experimental schema + generator written for the OpenSearch-Java client[1] that replicates what we intend to do for the python client. 

Another alternative would be to use the built in generator from either OpenAPI[2] or an accompanying python specific generator written for the OpenAPI spec[3]. This approach has drawbacks in the sense that they are not customizable to our needs. The current API layer is heavily opinionated in the way it is written, and using a generator from scratch would mean a lot of breaking changes in the function signatures, imports, tests, etc. We would like the initial approach to remain as close to the current API model as possible, and allow for a gradual overhaul of all the missing APIs and methods. Re-wiring the current generator allows us to do just that.   

[1] https://github.com/opensearch-project/opensearch-java/pull/366
[2] https://github.com/OpenAPITools/openapi-generator
[3] https://github.com/MarcoMuellner/openapi-python-generator

saimedhi:
### Description
Enhanced existing API generator to use the OpenAPI spec as the input format

### Issues Resolved
Closes #348 


VachaShah (commented):
Thank you @saimedhi!

For the implementation of the generator, lets not add the specs to the client. The generator should use the specs from the `opensearch-api-specification` repo. See https://github.com/opensearch-project/opensearch-api-specification/issues/74

dblock:
+1 for not checking in spec files and importing them, otherwise looks good!

nhtruong:
@VachaShah The OpenAPI specs on `opensearch-api-specification` repo currently doesn't have the extra metadata needed for generate the client code. This [PR](https://github.com/opensearch-project/opensearch-api-specification/pull/83) and its follow-up will have to be merged first before we can use the specs there for client generation.

Issue #393, opened 2023-05-15
[PROPOSAL] Split up test suites


dblock:
### What are you proposing?

On the tail of #392, split up test suites.

1. Unit tests that run by default and do _not_ require an instance of OpenSearch.
2. Integration tests that run against an instance or instances of OpenSearch.

### What users have asked for this feature?

Currently to get started in this project you must have docker and run an instance of OpenSearch. Many tests do not need to be going against the server and should be able to run offline. Those are unit tests and that's the suite that should run by default.

### What problems are you trying to solve?

Developers need too much infrastructure to get started.

### What is the developer experience going to be?

1. Check out source code.
2. Run `pipenv install`.
3. Run `pipenv run pytest`.

roma2023:
I will try to handle this issue. Can you please assign it to me?

dblock:
For anyone reading this and wanting to work on it, this issue drives me crazy. There are layers of `nox` and other `./ci` scripts wrapping nox that make no sense and require careful examination of the developer guide to figure out what to run. All that should be deleted in favor of simple package management, separate unit and integration tests, and all setup happening in GHA for various versions of Python or OpenSearch.

PR #394, opened 2023-05-16
[#387] Do not escape the "timeout" parameter.

### LINKED ISSUE #387: [BUG] Heathcheck timeout parameter not parsable
### What is the bug?

the OpenSearch client cannot change the default timeout for health checks.

The cluster endpoint for healthcheck requires a string representing time units:

```
http "os-morel-1:9200/_cluster/health?timeout=4s"
```

It fails if you supply an integer or float.

```
(rayish) √ deus_lex % http "os-morel-1:9200/_cluster/health?timeout=4.2"
HTTP/1.1 400 Bad Request
content-encoding: gzip
content-length: 172
content-type: application/json; charset=UTF-8

{
    "error": {
        "reason": "failed to parse setting [timeout] with value [4.2] as a time value: unit is missing or unrecognized",
        "root_cause": [
            {
                "reason": "failed to parse setting [timeout] with value [4.2] as a time value: unit is missing or unrecognized",
                "type": "illegal_argument_exception"
            }
        ],
        "type": "illegal_argument_exception"
    },
    "status": 400
}
```

However, the client call requires a numeric value and fails with a time unit string

```
x = OpenSearch(["http://os-morel-1:9200"])

x.cluster.health(timeout="4s")

ConnectionError: ConnectionError(Timeout value connect was b'4s', but it must be an int, float or None.) caused by: ValueError(Timeout value connect was b'4s', but it must be an int, float or None.)

x.cluster.health(timeout=4)

ConnectionError: ConnectionError(Timeout value connect was 4, but it must be an int, float or None.) caused by: ValueError(Timeout value connect was 4, but it must be an int, float or None.)
```

_Steps to reproduce the behavior._

As with previous versions of elasticsearch client, I expect "timeout="4s" to work properly

_Operating system, version._

_If applicable, add screenshots to help explain your problem._

_Add any other context about the problem._

grechaw:
When the "timeout" parameter is escaped and turned into a string, it does not pass validation checks.
This PR ensures that timeout is passed through as-is, just like "request_timeout".
There's no explicit test for this parameter, but I included it on some existing tests just to ensure that the code is exercised.  Without the fix in place, one of the tests as modified will not pass.

### Description
With this change, using the `timeout` parameter in `cluster.health` no longer throws a validation error.  I have not checked whether the problem exists elsewhere.

### Issues Resolved
Resolves #387 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -86,3 +86,4 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 [2.1.0]: https://github.com/opensearch-project/opensearch-py/compare/v2.0.1...v2.1.0
 [2.1.1]: https://github.com/opensearch-project/opensearch-py/compare/v2.1.0...v2.1.1
 [2.2.0]: https://github.com/opensearch-project/opensearch-py/compare/v2.1.1...v2.2.0
+
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -71,6 +71,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Updates `master` to `cluster_manager` to be inclusive ([#242](https://github.com/opensearch-project/opensearch-py/pull/242))
 - Support a custom signing service name for AWS SigV4 ([#268](https://github.com/opensearch-project/opensearch-py/pull/268))
 - Updated CI tests to make them work locally ([#275](https://github.com/opensearch-project/opensearch-py/pull/275))
+- Fix bug with validation of 'timeout' parameter ([#387](Do not escape the "timeout" parameter.))
 ### Deprecated
 
 ### Removed
--- a/opensearchpy/client/utils.py
+++ b/opensearchpy/client/utils.py
@@ -165,7 +165,7 @@ def query_params(*opensearch_query_params):
             elif api_key is not None:
                 headers["authorization"] = "ApiKey %s" % (_base64_auth_header(api_key),)
 
-            # don't treat ignore, request_timeout, timeout, and opaque_id as other params to avoid escaping
+            # don't escape ignore, request_timeout, or timeout
             for p in ("ignore", "request_timeout", "timeout"):
                 if p in kwargs:
                     params[p] = kwargs.pop(p)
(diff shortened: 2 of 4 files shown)


grechaw:
Based on feedback from my colleagues, I'll improve the testing.

dblock (changes requested):
- add to CHANGELOG, this fixes a bug I think
- you mention there's no unit test - try adding one that is scoped to the `utils` implementation for this?

grechaw:
Thanks for taking a look @dblock , I've added a simple test and I think this PR will help us move forward.  We have been moving to `opensearch-py` from a very old version of the Elasticsearch client.

CHANGELOG.md
```diff
@@ -71,6 +71,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Updates `master` to `cluster_manager` to be inclusive ([#242](https://github.com/opensearch-project/opensearch-py/pull/242))
 - Support a custom signing service name for AWS SigV4 ([#268](https://github.com/opensearch-project/opensearch-py/pull/268))
 - Updated CI tests to make them work locally ([#275](https://github.com/opensearch-project/opensearch-py/pull/275))
+- Fix bug with validation of 'timeout' parameter ([#387](Do not escape the "timeout" parameter.))
```
dblock, line 74: Re-add the space ;)

dblock (approved):
Great. This is good with me, but since you're here, extend this test to the other parameters (ignore and request_timeout)? I want to make sure we don't introduce regressions. You'll have to fix `nox lint` as well, please.

grechaw:
OK thanks again @dblock perhaps we are ready to go.   We all find the 'request_timeout' clobbering action odd, and I'm not sure what the 'ignore' parameter is for, but they're both under test now.

dblock (approved):

dblock:
LGTM! Thanks for hanging in here with me. 

Kicked integration tests again, we need to fix this flakiness ...

grechaw:
It's my pleasure.  I will keep my eye out for more things to address.  Where possible it will be good to migrate more of our code to the `opensearch-py` implementation.    I'll look forward to the next release.

dblock:
> It's my pleasure. I will keep my eye out for more things to address. Where possible it will be good to migrate more of our code to the `opensearch-py` implementation. I'll look forward to the next release.

If you want something meaty, I suggest https://github.com/opensearch-project/opensearch-py/issues/393 ;)

Issue #284, opened 2023-02-09
[FEATURE] Index State Management support


kinoute:
### Is your feature request related to a problem?

I would like to [create/add index policy](https://opensearch.org/docs/latest/im-plugin/ism/api/#create-policy) in the Index State Management but couldn't find a way to do it.

### What solution would you like?

Support The index state management to create policies.

### What alternatives have you considered?

I tried the official ElasticSearch which supports [the feature](https://elasticsearch-py.readthedocs.io/en/v7.17.9/api.html#elasticsearch.client.IlmClient.put_lifecycle) through the use of a `ilmClient` instance and the `put_lifecycle` method. But it doesn't work with OpenSearch because it tries to create policies in `/_ilm/policy/abc` which doesn't exist in OpenSearch:

```
no handler found for uri [/_ilm/policy/abc] and method [PUT]',
 'no handler found for uri [/_ilm/policy/abc] and method [PUT]')
```

wbeckler:
This is a good idea. Most of the plugins are unsupported via methods in the client, so developers end up writing raw curl requests and sending those like this (not in python):
```
PUT _plugins/_ism/policies/policy_id
{
  "policy": {
    "description": "Example policy.",
    "default_state": "...",
    "states": [...],
    "ism_template": {
      "index_patterns": ["index_name-*"],
      "priority": 100
    }
  }
}
```

There is a larger question of how to support plugins in clients. In OpenSearch, there is no such thing as an "ilmClient." 

I think one of the visions for OpenSearch is to be more extensible and flexible than its predecessor, which means it should be easy to add in this kind of functionality. And we should expect an increasing number of plugins/extensions.

Where do you think this functionality should go for the python client?

wbeckler:
I should also mention that someone started this conversation about how to architect this here: https://github.com/opensearch-project/opensearch-py/issues/90

kinoute:
For those who are interested, I forked the ElasticSearch Python library to support the creation of ISM/ISLM policies and it works:

https://github.com/kinoute/elasticsearch-py/blob/7.11/elasticsearch/client/ilm.py#L111-130

wbeckler:
Feel free to contribute any code to this repo that you write or find, as long as its license is compatible (e.g., Apache 2.0).

dblock:
Closed via https://github.com/opensearch-project/opensearch-py/pull/398

PR #398, opened 2023-05-25
Index State Management support

florianvazelle:
### Description
Added support for the ISM API.

### Issues Resolved
The issue of managing index states from the python client.  
This issue is the subject of a feature request at https://github.com/opensearch-project/opensearch-py/issues/284.


florianvazelle:
Hi @saimedhi, some tests seems to start opensearch without ISM plugin. Is this normal ?

https://github.com/opensearch-project/opensearch-py/actions/runs/5333494406/jobs/9664033038

saimedhi:
Hi @florianvazelle, I am not completely sure but I suggest to take a look at other existing plugin tests. I have gone through Alerting plugin tests. They are passing for "integration" tests and skipped for "Integration with Unreleased Opensearch" tests. So, please take a look at those tests once. Thanks :)

test_opensearchpy/test_server/test_plugins.py
```diff
@@ -181,3 +182,107 @@ def test_run_monitor(self):
         self.assertIn("monitor_name", response)
         self.assertIn("period_start", response)
         self.assertIn("period_end", response)
+
+
+class TestIndexManagementPlugin(OpenSearchTestCase):
```
dblock, line 187: We should break tests for each plugin in their own file(s), this is going to get really big.

dblock (changes requested):
Small request to move plugin tests into separate file(s). Please do for existing ones if this isn't the case.

dblock:
Looks good, needs a rebase + user guides.

florianvazelle:
@dblock done ! :slightly_smiling_face:

USER_GUIDE.md
```diff
@@ -21,6 +21,10 @@
       - [**Creating a destination**](#creating-a-destination)
       - [**Getting alerts**](#getting-alerts)
       - [**Acknowledge alerts**](#acknowledge-alerts)
```
dblock, line 23: Mind fixing these `**` in another PR? Thanks.

dblock (approved):

Issue #110, opened 2022-01-12
Missing SecurityClient - Inability to configure roles, role mappings and users


andyfase:
Seemingly the [ElasticSearch security client](https://github.com/elastic/elasticsearch-py/blob/main/elasticsearch/_sync/client/security.py) was deleted as part of a commit removing [xpack features ](https://github.com/opensearch-project/opensearch-py/commit/a41fd13d2d8a9bc8d800a53faedeb8e11d4cccd0). Though their maybe be xpack features in this client, it also contains methods that are needed to configure the following on AWS OpenSearch. Namely

- role methods (get, put, delete)
- user methods (get, put, delete)
- mapping methods  (get, put, delete)

Is there a plan to add this library back even in a reduced form? otherwise developers will be forced to revert. back to the elasticsearch-py lib for this kind of functionality.

Thanks

andyfase:
Note. I appreciate that I was wrong above, and the Security Client is the Native ES one and does not the OpenDistro ES client API endpoiints. 

Still the question remains if this library intends to release or would be interested in supporting core opendistro based methods for the types of methods mentioned above. Thanks.

anitakrueger:
I'd be interested in this as well, because it seems I cannot use the native Elasticsearch client on my OpenSearch cluster.
`elasticsearch7.exceptions.UnsupportedProductError: The client noticed that the server is not a supported distribution of Elasticsearch`
(using `elasticsearch7` python client on a `7.10.2` OpenDistro Elasticsearch.)

anitakrueger:
I got the native Elasticsearch client to work, but of course the SecurityClient doesn't work because of the difference in API endpoints.
How do I programmatically manage users and roles without making direct HTTP calls?

dblock:
This is similar to https://github.com/opensearch-project/opensearch-py/pull/93. That PR is in a good shape and just needs to be finished. Similarly we would need to add security plugin endpoints  here. Maybe someone wants to pick this up?

PR #399, opened 2023-05-29
Security plugin support

florianvazelle:
### Description
Added support for the Security API.

### Issues Resolved
The issue of managing roles, users... from the python client.  
This issue is the subject of a feature request at https://github.com/opensearch-project/opensearch-py/issues/110.


dblock (commented):
Looks good!

Can we add something to https://github.com/opensearch-project/opensearch-py/blob/main/USER_GUIDE.md#using-plugins as part of this PR?

florianvazelle:
I pushed some new commits to fix my tests that never passed, because I hadn't quite understood the test suite, which isn't very clear. I added the tests in an async version. 

Yes, I can add something to the USER_GUIDE, but the security API has a lot of enpoints, if I add examples for users and roles, is that enough?

.ci/run-opensearch.sh
```diff
@@ -81,7 +81,7 @@ END
     docker run \
       --name "$node_name" \
       --network "$network_name" \
-      --env "ES_JAVA_OPTS=-Xms1g -Xmx1g" \
+      --env "OPENSEARCH_JAVA_OPTS=-Xms1g -Xmx1g" \
```
dblock, line 84: Thanks for this.

CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added index lifecycle guide ([#362](https://github.com/opensearch-project/opensearch-py/pull/362))
 - Added 'point in time' APIs to the pyi files in sync and async client ([#378](https://github.com/opensearch-project/opensearch-py/pull/378))
 - Added MacOS and Windows CI workflows ([#390](https://github.com/opensearch-project/opensearch-py/pull/390))
+- Added support for the Security API ([#399](https://github.com/opensearch-project/opensearch-py/pull/399))
```
dblock, line 11: Let's say "security plugin"

USER_GUIDE.md
```diff
@@ -420,6 +425,66 @@ query = {
 response = client.plugins.alerting.acknowledge_alert(query)
 print(response)
 ```
+
+### Security plugin
+
+#### **Creating a role**
```
dblock, line 431: I think another PR was getting rid of these `**` bolding, let's remove those everywhere?

dblock (changes requested):
Looks great! Split out security plugin tests in its own file(s). Otherwise I think this is good enough.

dblock (approved):

saimedhi:
Hello @florianvazelle, can you please move Security Client from plugins to clients. example: (opensearchpy/client/security.py) like in PR https://github.com/opensearch-project/opensearch-py/pull/439. All the security features are considered as part of core features of OS. So, kindly let me know if you are willing to make the following changes. This is crucial and had to be done before next release. Thanks for your contributions.

PR #400, opened 2023-05-31
Fix integration tests workflow

florianvazelle:
### Description
The CI always seems to be using the latest version of opensearch rather than the one specified in the matrix.

image

### Issues Resolved
Fix integration tests that always use the latest version of opensearch rather than the one specified in the matrix.


dblock:
Ouch. Thanks.

PR #401, opened 2023-05-31
Added 'allow_redirects' parameter in perform_request function for RequestsHttpConnection

saimedhi:
### Description
Fixed CVE-2023-32681 linked to Requests

### Issues Resolved
fixes https://github.com/advisories/GHSA-j8r2-6x86-q33q

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added 'point in time' APIs to the pyi files in sync and async client ([#378](https://github.com/opensearch-project/opensearch-py/pull/378))
 - Added MacOS and Windows CI workflows ([#390](https://github.com/opensearch-project/opensearch-py/pull/390))
 - Compatibility with OpenSearch 2.1.0 - 2.6.0 ([#381](https://github.com/opensearch-project/opensearch-py/pull/381))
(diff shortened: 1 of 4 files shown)


harshavamsi:
Thanks @saimedhi for this change. Can you write a simple test on the `prepare_request` method mocking it and trying to do a redirect and failing(since `allow_redirects`) is false. This gives us confidence in this change working.

saimedhi:
PR not ready for review

dblock:
Thanks! Looks like this adds an `allow_redirects` option, aka it's a new feature?

1. Instead of calling out the CVE in CHANGELOG let's call out the new feature.
2. Document the new feature in user guide?

saimedhi:
Hello @dblock, Are you recommending that we need to capture 3xx response codes and ensure a new request is made to the redirect destination. This prevents breaking change for users currently relying on redirect behaviors. I agree with your suggestion and will implement it.

dblock:
Maybe, but I also think that you can start by turning this PR into just adding `allow_redirects` and defaulting it to `True`. We would merge that because it's just a new option, doesn't break anything. Then we can discuss the vulnerability and how we can either switch that option to `False` by default or implement something that can keep it at `True` and fix the vulnerability.

wbeckler:
If we merge it, it's important that we don't leave the impression that we "fixed" the CVE. That requires a default behavior of no auto-redirect. I'd be okay with 2 merges or 1 big one, but we need to get to the point where we default to the safe code path either way.

dblock (approved):

PR #419, opened 2023-06-20
Replaced botocore with aiobotocore in SigV4 asyncsigner

### LINKED ISSUE #410: [BUG] Blocking behavior in AWSV4SignerAsyncAuth
### What is the bug?

Yesterday's AWS us-east-1 outage revealed that botocore talks to STS using `requests`, and retries them using a backoff with `time.sleep`, which are both blocking functions and will cause performance degradation as they hold the event-loop hostage.

In this line: `sig_v4_auth.add_auth(aws_request)` https://github.com/opensearch-project/opensearch-py/blob/main/opensearchpy/helpers/asyncsigner.py#L54

Try calling a search operation using AWSV4SignerAsyncAuth while simulating poor network conditions to trigger the retry with backoff mechanism in botocore

It should not block the event-loop

Not relevant

image

Three possible solution proposals:
- Use aiobotocore, as it already takes care of that https://github.com/aio-libs/aiobotocore/blob/master/aiobotocore/signers.py
- Document a heads-up warning to use only frozen (read-only) credentials with AWSV4SignerAsyncAuth to prevent this blocking refresh operation
- Run the `sig_v4_auth.add_auth(aws_request)` call in a thread pool executor

saimedhi:
### Description
Replaced botocore with aiobotocore in SigV4 asyncsigner

### Issues Resolved
Closes #410 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Changed
 - Upgrading pytest-asyncio to latest version - 0.21.0 ([#339](https://github.com/opensearch-project/opensearch-py/pull/339))
 - Fixed flaky CI tests by replacing httpbin with a simple http_server ([#395](https://github.com/opensearch-project/opensearch-py/pull/395))
+- Replaced botocore with aiobotocore in SigV4 asyncsigner ([#417](https://github.com/opensearch-project/opensearch-py/pull/417))
 ### Deprecated
 ### Removed
 ### Fixed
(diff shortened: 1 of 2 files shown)


dacevedo12:
> Please let me know if any additional test cases are needed? 

the aiobotocore dependency needs to be declared.

It's not a drop-in replacement unfortunately. aiobotocore.auth does not exist

One would have to figure out how to use it with AioRequestSigner from aiobotocore

saimedhi:
> > Please let me know if any additional test cases are needed?
> 
> the aiobotocore dependency needs to be declared.
> 
> It's not a drop-in replacement unfortunately. aiobotocore.auth does not exist
> 
> One would have to figure out how to use it with AioRequestSigner from aiobotocore

@dacevedo12 , Thank you letting me know. I referred wrong documentation before. I will close this PR for now. Please give it a try if you find some time. Thanks :)

PR #424, opened 2023-06-27
Fix string/tuple/no auth on AsyncHttpConnection class

### LINKED ISSUE #283: [BUG] initializing an async client without http_auth results in an error

dannosaur:
### Description
Fixes an error and confusing code surrounding authentication on the AsyncHttpConnection class. aiohttp wants a `BasicAuth` instance instead of a tuple/colon-separated-string as requests does, so this accommodates what it wants also. The code in the `__init__` was lifted from the `RequestsHttpConnection` class as that's already a sensible implementation, but the way the resulting auth needs to be passed with the async session is slightly different.

### Issues Resolved
Closes #283


opensearchpy/connection/http_async.py
```diff
@@ -65,10 +65,13 @@ def __init__(
 
         if http_auth is not None:
             if isinstance(http_auth, (tuple, list)):
-                http_auth = ":".join(http_auth)
+                http_auth = tuple(http_auth)
             elif isinstance(http_auth, string_types):
                 http_auth = tuple(http_auth.split(":", 1))
 
+            if isinstance(http_auth, tuple):
+                http_auth = aiohttp.BasicAuth(login=http_auth[0], password=http_auth[1])
```
dblock, line 73: I think we are being too clever here. 

- We turn a tuple into a ... tuple. Seems unnecessary.
- We turn a list into a tuple.
- We turn a `string_types` into a tuple.
- Finally, we turn anything into `aiohttp.BasicAuth`.

Let's not be too clever, skip the intermediate `tuple` state and just make `aiohttp.BasicAuth` for any of the types?

```python
if isinstance(http_auth, tuple):
   http_auth = aiohttp.BasicAuth(login=http_auth[0], password=http_auth[1])
elif isinstance(http_auth, list)
   http_auth = aiohttp.BasicAuth(login=http_auth[0], password=http_auth[1])
elif isinstance(http_auth, string_types):
  http_auth = tuple(http_auth.split(":", 1))
```

opensearchpy/connection/http_async.py
```diff
@@ -190,17 +193,23 @@ async def perform_request(
             body = self._gzip_compress(body)
             req_headers["content-encoding"] = "gzip"
 
-        req_headers = {
-            **req_headers,
-            **self._http_auth(method, url, query_string, body),
-        }
+        auth = None
+        if self._http_auth:
+            if isinstance(self._http_auth, tuple):
```
dblock, line 198: This shouldn't be possible anymore because `__init__` turned everything into `aiohttp.BasicAuth`. Is that callable, too?

dblock (changes requested):
This needs tests for the input scenarios between tuple, strings, etc. Also is this the same problem not in async?

opensearchpy/connection/http_async.py
dannosaur: I tend to prefer DRY approaches, but that works too. I'll change it.

opensearchpy/connection/http_async.py
dannosaur: It _can_ be a callable if it's an instance of the AWS auth class, which doesn't belong in the `auth` parameter to the session, so it's covering that base.

This would simplify down to

```
auth = self._http_auth if isinstance(self._http_auth, aiohttp.BasicAuth) else None
```

We still need to maintain `auth=None` to the session if the AWS auth class was passed in.

opensearchpy/connection/http_async.py
dblock: Well, you choose what to DRY, `tuple` or `BasicAuth` :)

dannosaur: Fair lol

dannosaur:
> This needs tests for the input scenarios between tuple, strings, etc. Also is this the same problem not in async?

Sync mode works just fine. When I was switching my project over to async and it refused to authenticate I went digging.

dannosaur:
I'll push changes and tests for this tomorrow.

dblock:
> I'll push changes and tests for this tomorrow.

Thank you for helping out @dannosaur! No rush.

dannosaur:
Updated. One question. How do I omit my email from this required DCO? I don't want that out in the public domain.

dannosaur:
> How do I omit my email from this required DCO? I don't want that out in the public domain.

Nevermind. Found the Github stuff for hiding email addresses and rebased/did the DCO once again.

New to this :)

dblock:
> New to this :)

🤔 🦖 ;)

opensearchpy/connection/http_async.py
```diff
@@ -190,17 +191,20 @@ async def perform_request(
...
-        req_headers = {
-            **req_headers,
-            **self._http_auth(method, url, query_string, body),
-        }
+        auth = self._http_auth if isinstance(self._http_auth, aiohttp.BasicAuth) else None
+        if callable(self._http_auth):
+            req_headers = {
+                **req_headers,
+                **self._http_auth(method, url, query_string, body),
+            }
```
dblock, line 199: I believe `auth` is already `None` (if not it should be), isn't it, and these are mutually exclusive?

```
if isinstance(self._http_auth, aiohttp.BasicAuth):
   auth = self._http_auth
elif callable(self._http_auth):
   req_headers = {
     **req_headers,
     **self._http_auth(method, url, query_string, body),
   }
```

Not a must have, just thinking about making it more readable.

test_opensearchpy/test_connection.py
```diff
@@ -502,6 +504,46 @@ def urlopen_raise(*_, **__):
         assert str(e.value) == "Wasn't modified!"
 
 
+class TestAsyncHttpConnection(TestCase):
```
dblock, line 507: Let's pull this into a separate file? Either make an `async` folder with a bunch of tests or `test_async_http_connection.py`?

dblock (changes requested):
Add to CHANGELOG, and consider the other suggestions? LGTM otherwise.

opensearchpy/connection/http_async.py
dannosaur: `auth` is used in `self.session.request` a bit further down. Doing it this way will, when a callable is used for auth, raise a `NameError` because `auth` isn't defined. `auth=None` is the default to `session.request()` yes, but the only other alternative is to make 2 branches to call `self.session.request` with the different iterations, which I think is a lot messier. Doing it this way allows a single call to `session.request(auth=auth)`, where `auth` is `None` (the default), or an instance of `aiohttp.BasicAuth`.

test_opensearchpy/test_connection.py
dannosaur: Ok. Sure. I wasn't sure where to put it since it looked like this class didn't have a single test to its name :)

test_opensearchpy/test_async_http_connection.py
dannosaur, line 1: I noticed in coverage that the actual calls in `perform_request` hadn't been covered despite tests existing. Turns out async testing is a little more involved than what I'm used to. LMK if you want this done a different way. It seems the "official" way (as far as I can tell) on mocking out aiohttp.ClientSession is with a custom context manager mock class, coming from `asynctest` as opposed to the regular `mock` package.

dblock: I really don't know more than you about this, inherited this code :( Do what you think is the right way.

dblock:
Linter is complaining :( https://github.com/opensearch-project/opensearch-py/actions/runs/5427802648/jobs/9871401633?pr=424

I'm ok to merge it as is, lmk when it's ready and tests are passing.

test_opensearchpy/test_async_http_connection.py
dannosaur: In the interest of making future tests not have to import this context manager from this specific test file, I've moved it to the utils package.

dblock:
This looks good! Thanks for hanging in here. Linter/CI are still failing, will you please fix? 

I changed an option on this repo and hopefully CI will auto-run on this PR for you when you push updates.

dblock:
> Also, on the Python 3.7 front, it appears the IsolatedAsyncioTestCase class was added to unittest in Python 3.8. Is it acceptable to disable those tests on Python <3.8, or would you prefer that I explore options to have them run in all supported Python version?

Ideally all tests should run in all versions unless it's not possible (e.g. uses a feature of Python 3.8+). In that case I would accept an open issue and disabling those tests for <3.8 with a proper explanation.

dannosaur:
> Ideally all tests should run in all versions unless it's not possible (e.g. uses a feature of Python 3.8+). In that case I would accept an open issue and disabling those tests for <3.8 with a proper explanation.

Yeah, OK. I solved the 3.7 issue, but now 3.5 is complaining. I'll open a separate issue to talk about supported Python version before going any further with this.

dannosaur:
Woohoo!

@dblock I did open the issue about the older versions of Python, however I spotted that every other async-related test a) lives in the `test_async` directory, and b) is skipped for Python <3.6. So I moved the file, and it appears that the entire suite of checks is happy across the board. I simplified it somewhat as well, not relying on external libraries and pretty much copied the same format as the other tests were doing.

Let me know if you need anything else.

dblock (approved):

dblock:
Great job 👏

PR #425, opened 2023-06-29
Add async tests for plugins

florianvazelle:
### Description
1. Add asynchronous testing for plugins
2. Clean title in USER GUIDE for alerting section 

### Issues Resolved
1. To increase coverage
2. As requested here https://github.com/opensearch-project/opensearch-py/pull/398#discussion_r1245755900

### DIFF
--- a/test_opensearchpy/test_async/test_server/test_plugins/__init__.py
+++ b/test_opensearchpy/test_async/test_server/test_plugins/__init__.py
@@ -23,24 +23,3 @@
 #  KIND, either express or implied.  See the License for the
 #  specific language governing permissions and limitations
 #  under the License.
-
-
-from unittest import IsolatedAsyncioTestCase
-
-from opensearchpy._async.helpers.test import get_test_client
-from opensearchpy.connection.async_connections import add_connection
-
-from ...utils import wipe_cluster
-
-
-class AsyncOpenSearchTestCase(IsolatedAsyncioTestCase):
-    async def asyncSetUp(self):
-        self.client = await get_test_client(
-            verify_certs=False, http_auth=("admin", "admin")
-        )
-        await add_connection("default", self.client)
-
-    async def asyncTearDown(self):
-        wipe_cluster(self.client)
-        if self.client:
-            await self.client.close()
--- a/test_opensearchpy/test_async/test_server/test_plugins/test_alerting.py
+++ b/test_opensearchpy/test_async/test_server/test_plugins/test_alerting.py
@@ -13,9 +13,13 @@ from __future__ import unicode_literals
 
 import unittest
 
+import pytest
+
 from opensearchpy.helpers.test import OPENSEARCH_VERSION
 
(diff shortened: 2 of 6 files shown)


dblock:
LGTM, finish it up?

florianvazelle:
Hi @dblock, there are two points I'd like to look at:
- the tests pass but the teardown of the last test is flaky, because the event loop seems closed
- and coverage isn't increasing, because it seems that coverage reports sent to codecov are from the test workflow (https://github.com/opensearch-project/opensearch-py/blob/main/.github/workflows/test.yml), and in the test workflow no OpenSearch is run. So in fact coverage doesn't really represent what's being tested. I'm taking a little more time to see if it isn't possible to have a workflow that would launch all the tests (with a running OpenSearch), and that would send a complete report, which would make it possible to have a real coverage report on plugins for example.

dblock:
@florianvazelle I wouldn't worry about coverage

florianvazelle:
@dblock Ok, so It's good to me now

test_opensearchpy/test_async/test_server/test_plugins/test_alerting.py
```diff
@@ -0,0 +1,187 @@
+# -*- coding: utf-8 -*-
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+#
+# Modifications Copyright OpenSearch Contributors. See
```
dblock, line 8: You can remove the modifications part from new files. That was added during the fork for files that had an ES license before.

test_opensearchpy/test_async/test_server/test_plugins/__init__.py
```diff
@@ -0,0 +1,25 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+#
+# Modifications Copyright OpenSearch Contributors. See
```
dblock, line 7: This is a new file, so I believe everything including and down from Modifications ... should be removed. Just leave the SPDX header and the paragraph that says "The OpenSearch ... open source license.".

dblock (approved):
LGTM, two small changes with the license headers please? Sorry for the hassle.

tl;dr:

- new files only need the SPDX header + first paragraph
- existing files license headers should not be changed

test_opensearchpy/test_async/test_server/test_plugins/test_alerting.py
florianvazelle: But the `utils/license-headers.py` enforce the `Modifications Copyright OpenSearch Contributors` part. Should I adapt checks made in the script ?

dblock: Hmmm, we can deal with this separately.

PR #440, opened 2023-07-19
Add plugin clients typing to the opensearch client

florianvazelle:
### Description
I add the clients of the various plugins to the typing of the main client (`OpenSearch`).

### DIFF
--- a/opensearchpy/_async/client/__init__.pyi
+++ b/opensearchpy/_async/client/__init__.pyi
@@ -30,6 +30,9 @@ from __future__ import unicode_literals
 import logging
 from typing import Any, Collection, MutableMapping, Optional, Tuple, Type, Union
 
+from ..plugins.alerting import AlertingClient
+from ..plugins.index_management import IndexManagementClient
+from ..plugins.security import SecurityClient
 from ..transport import AsyncTransport
 from .cat import CatClient
 from .cluster import ClusterClient
@@ -56,6 +59,11 @@ class AsyncOpenSearch(object):
     remote: RemoteClient
     snapshot: SnapshotClient
     tasks: TasksClient
+
+    alerting: AlertingClient
+    security: SecurityClient
+    index_management: IndexManagementClient
+
     def __init__(
         self,
         hosts: Any = ...,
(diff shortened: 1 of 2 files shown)


dblock (commented):
Looks good, update CHANGELOG?

Is there anything we can do to test this in a way that we don't forget next time?

saimedhi:
Hello @florianvazelle, please respond to the following https://github.com/opensearch-project/opensearch-py/pull/399#issuecomment-1642789791. Thanks

saimedhi:
@florianvazelle, Could you please finish this PR when you have a moment? Thank you!

florianvazelle:
Hi @saimedhi, sorry I'm taking some time to study solutions to have a correct stubs package.

The only real way I've found is to use [stubtest](https://mypy.readthedocs.io/en/stable/stubtest.html), but that only ensures that the definitions in the stubs package are present and logical with what is defined in the package at runtime, but if a definition is missing in the stubs package it won't be raised.

I'm going to add this to the linting commands anyway.

To prevent oversights from happening again, I suggest opening a new issue to study solutions, and we'll can then merge this PR.

dblock:
@florianvazelle I finished #536 so types are back in the game. Do you want to finish this?

I am not sure we need to declare types like this here. For other types we just do `self.cat = CatClient(self)` and I think the Python type system is smart enough to figure out that `cat` is always `CatClient`, but I could be wrong. So either add types as proposed here or remove the declaration?

florianvazelle:
@dblock I think you fix my issue with #536, I'm closing this PR
thank you :wink:

dblock:
Did I? This code says `Any`, should it be typed or should the declaration be removed?

https://github.com/opensearch-project/opensearch-py/blob/d8dc5474b7e7e2b443d9858c21d8f7be93306704/opensearchpy/client/plugins.py#L21

dblock:
Did I? This code says `Any`, should it be typed or should the declaration be removed?

https://github.com/opensearch-project/opensearch-py/blob/d8dc5474b7e7e2b443d9858c21d8f7be93306704/opensearchpy/client/plugins.py#L21

PR #442, opened 2023-07-20
Move security api from plugins to clients

florianvazelle:
### Description
I move the `SecurityClient` from the plugins directory to the clients directory, accordingly to the @saimedhi comment (https://github.com/opensearch-project/opensearch-py/pull/399#issuecomment-1642789791).
> All the security features are considered as part of core features of OS.


dblock:
How does this impact end-users? Anything I'd have to change in my code when upgrading?

dblock:
> You'll just have to change the import paths if you imported the client directly from plugins (`from opensearchpy.plugins import security` into `from opensearchpy.clients import security`).

That's a breaking change :( Is there a way to proxy the old to the new so we don't have to bump the major version?

saimedhi:
> > You'll just have to change the import paths if you imported the client directly from plugins (`from opensearchpy.plugins import security` into `from opensearchpy.clients import security`).
> 
> That's a breaking change :( Is there a way to proxy the old to the new so we don't have to bump the major version?

@dblock This change is not yet released.

dblock:
> > > You'll just have to change the import paths if you imported the client directly from plugins (`from opensearchpy.plugins import security` into `from opensearchpy.clients import security`).
> > 
> > 
> > That's a breaking change :( Is there a way to proxy the old to the new so we don't have to bump the major version?
> 
> @dblock This change is not yet released.

Oh! Good.

dblock (approved):

Issue #435, opened 2023-07-14
[FEATURE] Add support for server-side APIs that are currently not implemented in the client


saimedhi:
### What solution would you like?
Add support for server-side APIs that are currently not implemented in the client

dblock:
Since we added code generation from spec I am going to close this. We have issues for specific APIs missing under the [plugin support](https://github.com/opensearch-project/opensearch-py/issues?q=is%3Aopen+is%3Aissue+label%3A%22plugin+support%22) label for example for plugins, so anyone reading this please open a new issue for any specific API missing.

PR #443, opened 2023-07-20
Added restore API in RemoteStoreClient

saimedhi:
### Description
Added restore API in RemoteStoreClient. This is an experimental feature. API Code is generated using "python API generator". Tests to this feature can be added [here](https://github.com/opensearch-project/OpenSearch/tree/main/rest-api-spec/src/main/resources/rest-api-spec/test)

### Issues Resolved
Related Issue https://github.com/opensearch-project/opensearch-py/issues/435


dblock:
The code is generated but shouldn't we have _some_ tests for it to make sure it works at all? Maybe we can also take the habit of documenting this as part of these PRs in the user guides/adding working samples?

PR #444, opened 2023-07-20
Enhanced API generator to generate imports and headers

saimedhi:
### Description
Enhanced API generator to generate imports and headers

### Issues Resolved
Closes https://github.com/opensearch-project/opensearch-py/issues/348

### DIFF
--- a/utils/generate-api.py
+++ b/utils/generate-api.py
@@ -98,7 +98,14 @@ class Module:
 
     def parse_orig(self):
         self.orders = []
-        self.header = "class C:"
+        self.header = ""
+        if self.is_pyi is True:
+            self.header = "from typing import Any, Collection, MutableMapping, Optional, Tuple, Union\n\n"
+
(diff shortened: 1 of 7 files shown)


utils/generate-api.py
```diff
@@ -130,11 +137,136 @@ def sort(self):
...
+                f.write(self.header[License_position:])
+            else:
+                f.write(self.header[:header_position])
+                f.write("\n")
+                f.write("#replace_token#\n")
+                f.write(self.header[header_position:])
             for api in self._apis:
                 f.write(api.to_python())
 
+        # Added "delete point in time" API manually to prevent breaking changes
```
dblock, line 205: Yikes, this seems like a huge hack. We need to find a better way, for example by putting generated APIs into `generated/` and adding `patches/` folder that includes those APIs that should not be generated. You can use file names to decide which one needs to be which.

dblock (changes requested):
This is way too hacky ;) Suggestions below?

PR #470, opened 2023-08-10
Fix race condition in AWS request signers

### LINKED ISSUE #423: [BUG] AWSV4SignerAuth doesn't freeze credentials before signing, leading to a race condition
### What is the bug?
AWSV4SignerAuth passes the raw credentials object to SigV4Auth, instead of checking for a `get_frozen_credentials` method and calling that. This leads to the access key for a request differing from the secret key used to sign it.

1. Setup your AWS SDK with a refreshable credential mechanism.
2. Sign requests like it's going out of business
3. When you cross the refresh threshold, you'll get a request indicating it was signed with a different key than it actually was.

All requests get signed with matching access keys & secrets.

See this Python docstring in botocore for a more thorough explanation of the problem.

https://github.com/boto/botocore/blob/c29640a5e7a5f77cc4c6bb27e30a5387bc12832a/botocore/credentials.py#L399-L406

lattwood:
Closes #423

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Deprecated
 ### Removed
 ### Fixed
+- Fixed race condition in AWSV4SignerAuth & AWSV4SignerAsyncAuth when using refreshable credentials ([#470](https://github.com/opensearch-project/opensearch-py/pull/470))
 ### Security
 ### Dependencies
 - Bumps `sphinx` from <7.1 to <7.2
--- a/test_opensearchpy/test_async/test_signer.py
+++ b/test_opensearchpy/test_async/test_signer.py
@@ -18,7 +18,7 @@ pytestmark = pytest.mark.asyncio
 
 
 class TestAsyncSigner:
-    def mock_session(self, disable_get_frozen=True):
+    def mock_session(self):
         access_key = uuid.uuid4().hex
         secret_key = uuid.uuid4().hex
         token = uuid.uuid4().hex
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -115,4 +116,4 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 [2.1.0]: https://github.com/opensearch-project/opensearch-py/compare/v2.0.1...v2.1.0
 [2.1.1]: https://github.com/opensearch-project/opensearch-py/compare/v2.1.0...v2.1.1
 [2.2.0]: https://github.com/opensearch-project/opensearch-py/compare/v2.1.1...v2.2.0
-[2.3.0]: https://github.com/opensearch-project/opensearch-py/compare/v2.2.0...v2.3.0
\ No newline at end of file
+[2.3.0]: https://github.com/opensearch-project/opensearch-py/compare/v2.2.0...v2.3.0
```
lattwood, line 119: This showed up because my local `vim` doesn't like it when a file doesn't end in a newline- I'm guessing most the devs on this project are using VSCode

dblock: It's fine.

test_opensearchpy/test_async/test_signer.py
```diff
@@ -18,19 +18,38 @@
...
         access_key = uuid.uuid4().hex
         secret_key = uuid.uuid4().hex
         token = uuid.uuid4().hex
         dummy_session = Mock()
         dummy_session.access_key = access_key
         dummy_session.secret_key = secret_key
         dummy_session.token = token
+        dummy_session.get_frozen_credentials = Mock(return_value=dummy_session)
+
+        if disable_get_frozen:
```
dblock, line 31: There's probably a cleaner way to do this, if you care, by inheriting `TestAsyncSigner` into a `TestAsyncSignerWithFrozenCredentials(TestAsyncSigner)` or something like that?

dblock (commented):
Looks good!

test_opensearchpy/test_async/test_signer.py
lattwood: Hrm. The issue is the checks for the `get_frozen_credentials` method in the signer implementations will always evaluate to `True` if I don't explicitly delete the attribute because of how eager `Mock` is to mock :)

lattwood:
@dblock unless you can think of a better solution to `Mock` being overeager around the `get_frozen_credentials` method (that I'd be happy to implement), I'd say go ahead and merge it. Anything I need to do for backports?

lattwood:
Figured it out, another commit will be forthcoming

lattwood:
doing one more force push to retrigger that failing integration test

dblock:
> doing one more force push to retrigger that failing integration test

This failure has been so annoying. Hope someone can make the test more reliable. Wink wink.

dblock (approved):

lattwood:
> > doing one more force push to retrigger that failing integration test
> 
> This failure has been so annoying. Hope someone can make the test more reliable. Wink wink.

Looks like it's related to docker networking stuff 😅

dblock:
> Anything I need to do for backports?

AFAIK no, we release from main, @VachaShah yes?

lattwood:
> > Anything I need to do for backports?
> 
> AFAIK no, we release from main, @VachaShah yes?

image

was why I asked (CONTRIBUTING.md)

dblock:
Good work @lattwood !

VachaShah:
> > Anything I need to do for backports?
> 
> AFAIK no, we release from main, @VachaShah yes?

Yeah no backports are required currently as we release from main since there have been no breaking changes in this client, so main = 2.x.

PR #483, opened 2023-08-22
Revised Security Client APIs to achieve consistency with other clients

saimedhi:
### Description
_Describe what this change achieves._

### Issues Resolved


saimedhi:
To achieve consistency with cat client, Renamed "health_check" API to "Health".  To achieve consistency with other language clients and also other similar APIs, Renamed "update_audit_config" API to "update_audit_configuration".

dblock (commented):
I see a bunch of breaking changes here. Is this the first time we're shipping security client support? If not we probably need to keep old names that call new names.

dblock:
Is this still needed with client code generation?

Issue #184, opened 2022-08-05
[BUG] Allow setting alternate host header (with port) in AWSV4SignerAuth 


daniel-rhoades:
### What is the bug?
The [`fetch_url` method within `signer.py`](https://github.com/opensearch-project/opensearch-py/blob/035b47265045409d5687a878dbbfdd99894af25c/opensearchpy/helpers/signer.py#L42) does not include port information in the signing request - it only specifies uses the `host` field.

As a result, the signing request is incorrectly generated when connecting to an OpenSearch domain through a proxy or tunnel which uses a different port (e.g. 8443) than the domain is configured for (e.g. 443).  This causes an `opensearchpy.exceptions.AuthorizationException` when attempting to make requests as the expected URL (without a port number) given to the signer doesn't match the request being made (which has the port number).

### How can one reproduce the bug?
Using either: 

* An OpenSearch domain secured by AWS IAM; or more specifically 
* Isolated AWS OpenSearch domain (e.g. deploy into a VPC without an IGW, relying on the appropriate VPC Endpoints instead)

1. Establish a proxy (e.g. a local Nginx proxy) or tunnel (e.g. SSH tunnel to a bastion host) to the domain's endpoint but use a _different port_ from the domain's default port (e.g. use 8443 instead of 443)
2. Ensure the hostname will correctly resolve to the proxy or tunnel, either:
    a. Add an entry to the local hosts file; or
    b. Use `localhost` as the host name below and set `verify_certs` to `False`
3. Follow the documented example for connecting to and querying an index, but ensure the client is configured to use the port of the local proxy / tunnel (see example below)

Replace `xxx` as appropriate for `region`, `host` and `index`.  The code assumes appropriate AWS credentials are already available within the environment this code is executed.

```
from opensearchpy import OpenSearch, RequestsHttpConnection, AWSV4SignerAuth
import boto3

region = 'xxx'
host = f'xxx.{region}.es.amazonaws.com'
port = 8443

credentials = boto3.Session().get_credentials()
auth = AWSV4SignerAuth(credentials, region)

client = OpenSearch(
    hosts=[{'host': host, 'port': port}],
    http_auth=auth,
    use_ssl=True,
    verify_certs=False,
    connection_class=RequestsHttpConnection
)

client.search(
    body=None,
    index='xxx'
)
```

### What is the expected behavior?
[...]

PR #491, opened 2023-09-12
feat: allow to set a signature port for tunnel usage

andreaslang:
### Description
Allows to change the port used to sign the AWS request which is causing an issue if accessing an AWS Opensearch instance via a tunnel

If you have an ssh tunnel created this works now (while it would not without `signature_port`):
```
host = 'localhost'
port = 10012
region = 'eu-west-1'
service = 'es'
auth = AWSV4SignerAuth(credentials, region, service, signature_port=443)

client = OpenSearch(
    hosts = [{'host': host, 'port': port}],
    http_auth = auth,
    use_ssl = True,
    verify_certs = False,
    ssl_assert_hostname = False,
    ssl_show_warn = False,
    connection_class = RequestsHttpConnection,
    pool_maxsize = 20
)
```

### Issues Resolved
https://github.com/opensearch-project/opensearch-py/issues/184

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Deprecated
 ### Removed
 ### Fixed
+- Fixed allow AWSV4SignerAuth to work with a tunnel ([[#184](https://github.com/opensearch-project/opensearch-py/issues/184)
 ### Security
 ### Dependencies
 - Bumps `sphinx` from <7.1 to <7.3
--- a/opensearchpy/helpers/signer.py
+++ b/opensearchpy/helpers/signer.py
@@ -6,7 +6,6 @@
 #
 # Modifications Copyright OpenSearch Contributors. See
 # GitHub history for details.
-
 import sys
 
 import requests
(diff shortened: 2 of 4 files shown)


dblock (changes requested):
This is an interesting workaround, but I really don't think it's the right one. Using a tunnel is already quite suspicious for auth, it feels like MITM. With this change we're basically telling the signer to use a different host/port in some very special case - I am struggling to convince myself it's a good idea.

1. What do other clients do? Do they break the same way?
2. Can we just document this case and use a slightly more custom `AWSV4SignerAuth`?

dblock:
I would merge a change that allows to override headers, and specifically to override the `host` header.

andreaslang:
> I would merge a change that allows to override headers, and specifically to override the `host` header.

Yes, that is a really good point. It should work the same way if I replace the host header like localhost:10012 > realdomain:443. I will change the PR having an Authentication class specifically for that (e.g. `TunneledAWSV4SignerAuth`) with the explanation in the docstring. Will likely get to it next Tuesday.

dblock:
> > I would merge a change that allows to override headers, and specifically to override the `host` header.
> 
> Yes, that is a really good point. It should work the same way if I replace the host header like localhost:10012 > realdomain:443. I will change the PR having an Authentication class specifically for that (e.g. `TunneledAWSV4SignerAuth`) with the explanation in the docstring. Will likely get to it next Tuesday.

I actually would prefer as a developer to be able to write something like this:

```python
client = OpenSearch(
    hosts = [{'host': host, 'port': port}],
    http_headers: {
        'Host': 'something.us-west2.aws.bla.bla.bla'
    }
)
```

The signer implementation should be smart enough to consider this `host` header. 

This seems more generic and future-proof, don't you think?

dblock:
@andreaslang Are you still interested in adding an `http_headers` option?

andreaslang:
Yes, I am. Sorry for the delay. I do have a good excuse that my daughter
was born though. Planning to pick this up soon.

saimedhi:
Hello @andreaslang! Hope you're doing great. Could you please consider finishing up this PR when you have a moment? Thank you!

dblock:
> Hi! I'm interested in getting this one sorted and am happy to contribute to this MR and implement the `http_headers` option. Let me know if I can go ahead, please.

Please do!

brunomurino:
I started looking into this and I believe your suggestion is already somewhat implemented and satisfies this issue.
This is what I used to connect to OpenSearch behind a Bastion server:
```
hosts=[{
        "host": "<aws opensearch endpoint>", -- something like 'foo.eu-west-1.es.amazonaws.com', I did add this to my local /etc/hosts to pass the hostname check, which is trivial
        "port": 3401, -- port I opened the SSM/SSH tunnel on
        "headers": {
            'Host': '<aws opensearch endpoint>:443' -- the real endpoint
        }
    }],
```
Without passing the "headers" I was getting a connection refused because the request signature wasn't what it was expected due to the port being different, and after passing the "headers" option I did manage to connect successfully.

Correct me if I'm wrong, but no changes are required at least since version 2.7.1

dblock:
@brunomurino Check that this works with all the transport classes we have, maybe write some tests if any are missing, and update docs, and we can close this?

PR #520, opened 2023-10-03
fixes: #512 Wrong return type hint in async_scan

### LINKED ISSUE #512: [BUG] Wrong return type hint in async_scan
### What is the bug?

The interface for `async_scan` (https://github.com/opensearch-project/opensearch-py/blob/2feccc22bb8300e9e0d9a4575e2717f2baec17ec/opensearchpy/_async/helpers/actions.pyi#L103C6-L103C31) suggests that the return type is `AsyncGenerator[int, None]`, while it should be `AsyncGenerator[dict, None]` since every yielded hit is a dictionary.

bhavaniravi:
### Description
fixes: #512 Wrong return type hint in async_scan
Verified the same with [opensearch API](https://opensearch.org/docs/latest/search-plugins/async/index/)

### Issues Resolved
Closes #512 


dblock:
Is there a way we can test this? I'm thinking if there was a unit test that would exercise this API, then `mypy` would fail on it?

dblock (approved):

PR #535, opened 2023-10-12
Align pool_maxsize for different connection pool implementations.

dblock:
### Description

- Enable `pool_maxsize` for both `Urllib3HttpConnection` and `RequestsHttpConnection` pool implementations.
- Document how and when to switch between connection pools.
- Remove `RequestsHttpConnection` from the documentation when used as a default.

### DIFF
--- a/guides/connection_classes.md
+++ b/guides/connection_classes.md
@@ -7,7 +7,7 @@
 
 # Connection Classes
 
-The OpenSearch Python synchrnous client supports both the `Urllib3HttpConnection` connection class (default) from the [urllib3](https://pypi.org/project/urllib3/) library, and `RequestsHttpConnection` from the [requests](https://pypi.org/project/requests/) library. We recommend you use the default, unless your application is standardized on `requests`. 
+The OpenSearch Python synchronous client supports both the `Urllib3HttpConnection` connection class (default) from the [urllib3](https://pypi.org/project/urllib3/) library, and `RequestsHttpConnection` from the [requests](https://pypi.org/project/requests/) library. We recommend you use the default, unless your application is standardized on `requests`.
 
 The faster, asynchronous client, implements a class called `AsyncHttpConnection`, which uses [aiohttp](https://pypi.org/project/aiohttp/).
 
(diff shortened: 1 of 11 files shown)


guides/connection_classes.md
```diff
@@ -0,0 +1,81 @@
+- [Connection Classes](#connection-classes)
+  - [Selecting a Connection Class](#selecting-a-connection-class)
+    - [Urllib3HttpConnection](#urllib3httpconnection)
+    - [RequestsHttpConnection](#requestshttpconnection)
+    - [AsyncHttpConnection](#asynchttpconnection)
+  - [Connection Pooling](#connection-pooling)
+
+# Connection Classes
+
+The OpenSearch Python synchrnous client supports both the `Urllib3HttpConnection` connection class (default) from the [urllib3](https://pypi.org/project/urllib3/) library, and `RequestsHttpConnection` from the [requests](https://pypi.org/project/requests/) library. We recommend you use the default, unless your application is standardized on `requests`.
```
VachaShah, line 10: *Nit*: synchronous

guides/connection_classes.md
```diff
@@ -0,0 +1,81 @@
...
+```python
+from opensearchpy import OpenSearch, Urllib3HttpConnection
+
+client = OpenSearch(
+    hosts = [{'host': 'localhost', 'port': 9200}],
+    http_auth = ('admin', 'admin'),
+    use_ssl = True,
+    verify_certs = False,
+    ssl_show_warn = False,
+    connection_class = Urllib3HttpConnection
```
VachaShah, line 27: Should we make `Urllib3HttpConnection` default in code unless a `connection_class` is explicitly passed in?

dblock: It's already default here: https://github.com/opensearch-project/opensearch-py/blob/84ac172ddc54b3e6c975d36221d16ec3e78a2fe9/opensearchpy/transport.py#L70

guides/connection_classes.md
dblock: Good 👀

guides/connection_classes.md
VachaShah: Ah I missed it, so only the documentation was a bit misleading.

VachaShah (approved):
Thank you @dblock!

navneet1v:
@dblock can we add the recommendation somewhere in doc on what HTTPClient to use? with the benchmarks you have run.

dblock:
> @dblock can we add the recommendation somewhere in doc on what HTTPClient to use? with the benchmarks you have run.

In this PR I had added "We recommend you use the default, unless your application is standardized on `requests`." LMK if you think it should say something else.

Issue #477, opened 2023-08-16
[PROPOSAL]Merge Generated APIs into Existing Client Module


saimedhi:
### What are you proposing?
Merge the generated APIs into the existing client module to progress towards complete client generation, ensuring alignment with the server.

### How does the opensearch-py API [Generator](https://github.com/opensearch-project/opensearch-py/blob/main/utils/generate-api.py) function?
1. Run the generator as per the [developer guide](https://github.com/opensearch-project/opensearch-py/blob/main/DEVELOPER_GUIDE.md#running-python-client-generator).
2. The generator retrieves the [OpenSearch API specification](https://github.com/opensearch-project/opensearch-api-specification/blob/main/OpenSearch.openapi.json) JSON data.
3. API paths and methods are processed and grouped based on the x-operation-group field.
4. API parameters, parts, methods, and documentation are updated and organized.
5. For each API endpoint, instances of Module and API classes are initialized.
6. Generator checks if corresponding module exists in the clients folder for each API's namespace.
7. If the module is absent, creates a module for the corresponding namespace.
8. In the module, a class is created based on the namespace, containing API definitions or actions.
9. Ninja templates are utilized for generation, with overrides for compatibility.
10. Format the code using instructions in developer guide. 
11. If a namespace/module already exists, then existing definitions in that module are replaced by the generated API definitions.

### What problems are you trying to solve?

 Consider 'indices' namespace, post-generation, existing APIs are replaced by generated ones. Some generated API may be different from the existing APIs. We need to explore solutions for handling divergent APIs, termed as 'Patch APIs.' Our goal is to structure the merging process to prevent breaking changes. We categorize APIs as 'generatable' or 'Patch'

<img width="485" alt="image" src="https://github.com/opensearch-project/opensearch-py/assets/117196660/3cc33190-ed93-4339-a235-0f8fed03320a">

### Solutions:
Three approaches are discussed below to separate and structure generatable and patch APIs.

#### Current client structure:
<img width="447" alt="image" src="https://github.com/opensearch-project/opensearch-py/assets/117196660/1bbfbdc7-b2dc-45c9-bc08-91e893bb162d">

#### Approach 1
Consider "indices" namespace, all the patch APIs are placed in a separate file "indices/patch.py". In this way while running generator, patch APIs wont be altered. And we generate all other APIs into "indices.py". Easy implementation and contributor-friendly.

[...]

PR #543, opened 2023-10-16
Generate opensearch-py client from API specs

saimedhi:
### Description
Generate opensearch-py client from API specs

### Issues Resolved
Related to #477 

### DIFF
--- a/opensearchpy/_async/client/ingest.py
+++ b/opensearchpy/_async/client/ingest.py
@@ -117,7 +117,7 @@ class IngestClient(NamespacedClient):
         :arg body: The simulate definition
         :arg id: Pipeline ID.
         :arg verbose: Verbose mode. Display data output for each
-            processor in executed pipeline.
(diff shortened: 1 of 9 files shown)


opensearchpy/_async/client/ingest.py
```diff
@@ -117,7 +117,7 @@ async def simulate(self, body, id=None, params=None, headers=None):
         :arg body: The simulate definition
         :arg id: Pipeline ID.
         :arg verbose: Verbose mode. Display data output for each
-            processor in executed pipeline.
+            processor in executed pipeline. (default: false)
```
dblock, line 120: Nit: when we update templates we can say "Default is false." so it reads better.

dblock:
@saimedhi Anything here that you didn't want to merge? I got it to green in https://github.com/opensearch-project/opensearch-py/pull/549.

PR #434, opened 2023-07-13
Generated python client

saimedhi:
### Description
Generated python client

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #549, opened 2023-10-24
Use API generator.

dblock:
### Description

Finishes #434 and https://github.com/opensearch-project/opensearch-py/pull/543.

Introduces a way to override API spec.

### Issues Resolved


dblock:
@saimedhi When running the generator we end up with 1 failing test that calls `put_mapping` with an index name that is not optional extracted from the path. Looks like OpenSearch supports `_all`, and the Python code gladly facilitates that by overriding the implementation and setting the value. We also need to override the method definition so that `index = None`. I couldn't find a way to do it, so I came up with patching the OpenAPI spec. Not sure whether this is the right approach?

saimedhi:
> @saimedhi When running the generator we end up with 1 failing test that calls `put_mapping` with an index name that is not optional extracted from the path. Looks like OpenSearch supports `_all`, and the Python code gladly facilitates that by overriding the implementation and setting the value. We also need to override the method definition so that `index = None`. I couldn't find a way to do it, so I came up with patching the OpenAPI spec. Not sure whether this is the right approach?

Thank you @dblock, for working on this. Can I take it from here? For put_mapping, alternative approach is to use the below code in the generator.

```
        if namespace == "indices" and name == "put_mapping":
            api["url"]["paths"][0]["parts"]["index"].update({"required": False})
```

dblock:
> > @saimedhi When running the generator we end up with 1 failing test that calls `put_mapping` with an index name that is not optional extracted from the path. Looks like OpenSearch supports `_all`, and the Python code gladly facilitates that by overriding the implementation and setting the value. We also need to override the method definition so that `index = None`. I couldn't find a way to do it, so I came up with patching the OpenAPI spec. Not sure whether this is the right approach?
> 
> Thank you @dblock, for working on this. Can I take it from here? For put_mapping, alternative approach is to use the below code in the generator.
> 
> ```
>         if namespace == "indices" and name == "put_mapping":
>             api["url"]["paths"][0]["parts"]["index"].update({"required": False})
> ```

Yes of course. I think my version of patching the API definition is a bit more generic :)

dblock:
Will close this one, feel free to take any parts of it @saimedhi into #543.

PR #553, opened 2023-10-26
Override put_mapping definition to make index name optional.

dblock:
### Description

This is an alternative to patching in code.

### DIFF
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -7,6 +7,7 @@ sphinx<7.3
 sphinx_rtd_theme
 jinja2
 pytz
+deepmerge
 
 # No wheels for Python 3.10 yet!
 numpy; python_version<"3.10"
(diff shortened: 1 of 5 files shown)


opensearchpy/client/plugins.py
```diff
@@ -45,9 +44,7 @@ def _dynamic_lookup(self, client):
                 setattr(client, plugin, getattr(self, plugin))
             else:
                 warnings.warn(
-                    "Cannot load `{plugin}` directly to OpenSearch. `{plugin}` already exists in OpenSearch. Please use `OpenSearch.plugin.{plugin}` instead.".format(
-                        plugin=plugin
-                    ),
+                    f"Cannot load `{plugin}` directly to AsyncOpenSearch. `{plugin}` already exists in AsyncOpenSearch. Please use `AsyncOpenSearch.plugin.{plugin}` instead.",
```
saimedhi, line 47: The generator has replacing **AsyncOpenSearch** with **OpenSearch** code, but it's not happening, and I'm not sure why. https://github.com/saimedhi/opensearch-py/blob/main/utils/generate-api.py#L655

saimedhi (changes requested):

opensearchpy/client/plugins.py
dblock: I'll look.

dblock:
@saimedhi please don't hesitate to hit merge if you approve of a PR and it passes (at least from me!)

Issue #582, opened 2023-11-14
[BUG] Integration test failures against 2.11


dblock:
### What is the bug?

```
FAILED test_opensearchpy/test_server/test_rest_api_spec.py::test_rest_api_spec[OpenSearch-main/rest-api-spec/src/main/resources/rest-api-spec/test/search/260_sort_mixed[2]]
FAILED test_opensearchpy/test_server/test_rest_api_spec.py::test_rest_api_spec[OpenSearch-main/rest-api-spec/src/main/resources/rest-api-spec/test/search/90_search_after[4]]
```

```
E       opensearchpy.exceptions.RequestError: RequestError(400, 'search_phase_execution_exception', 'Field population is indexed with 2 bytes per dimension, but org.opensearch.index.fielddata.fieldcomparator.FloatValuesComparatorSource$1@2570133f expected 4')
```

```
DEBUG    opensearch:base.py:200 < {"error":{"root_cause":[{"type":"illegal_argument_exception","reason":"Field population is indexed with 2 bytes per dimension, but org.opensearch.index.fielddata.fieldcomparator.FloatValuesComparatorSource$1@2570133f expected 4"}],"type":"search_phase_execution_exception","reason":"all shards failed","phase":"query","grouped":true,"failed_shards":[{"shard":0,"index":"test","node":"EMIK3CDxT5qq2xP112An6A","reason":{"type":"illegal_argument_exception","reason":"Field population is indexed with 2 bytes per dimension, but org.opensearch.index.fielddata.fieldcomparator.FloatValuesComparatorSource$1@2570133f expected 4"}}],"caused_by":{"type":"illegal_argument_exception","reason":"Field population is indexed with 2 bytes per dimension, but org.opensearch.index.fielddata.fieldcomparator.FloatValuesComparatorSource$1@2570133f expected 4","caused_by":{"type":"illegal_argument_exception","reason":"Field population is indexed with 2 bytes per dimension, but org.opensearch.index.fielddata.fieldcomparator.FloatValuesComparatorSource$1@2570133f expected 4"}}},"status":400}
```

https://github.com/opensearch-project/opensearch-py/actions/runs/6868286935/job/18678495702

### How can one reproduce the bug?

```
./.ci/run-tests false 2.11.0 test_rest_api_spec
```

### What is the expected behavior?

Tests to pass.

### What is your host/environment?

CI/CD 2.11.

### Do you have any additional context?

[...]

PR #583, opened 2023-11-14
Remove 2.11.

dblock:
### Description

Removing 2.11 because of #582. I am 99.9% sure this is not a client problem, see https://github.com/opensearch-project/OpenSearch/pull/11024#issuecomment-1811264963.


VachaShah (approved):
Do we also want to remove it from COMPATIBILITY.md?

dblock:
> Do we also want to remove it from COMPATIBILITY.md?

No I think the test is wrong, but I don't know how to fix it.

PR #588, opened 2023-11-15
Fix: version number in filename for distribution.

dblock:
### Description

Looks like we released https://pypi.org/project/opensearch-py/0.0.0/ 😢 


utils/build-dists.py
```diff
@@ -258,8 +258,8 @@ def main() -> None:
         with open(version_path) as f:
             version_data = f.read()
         version_data = re.sub(
-            r"__versionstr__ = \"[^\"]+\"",
-            '__versionstr__ = "%s"' % version,
+            r"__versionstr__: str = \"[^\"]+\"",
+            '__versionstr__: str = "%s"' % version,
```
dblock, line 262: TBH I don't understand what any of this code is trying to do.

VachaShah (approved):
Interesting! I have no idea what this does or why 🤔

utils/build-dists.py
```diff
@@ -194,7 +194,7 @@ def main() -> None:
         if m:
             version = m.group(1)
         else:
-            raise Exception(f"Invalid version {data}")
+            raise Exception(f"Invalid version: {data}")
```
VachaShah, line 197: Did we get this exception? Or did the code just change the version to 0.0.0?

dblock: I just changed the text to match another place we throw it.

PR #598, opened 2023-11-16
tasks.py: warn if task_id IS skipped

aviau:


aviau:
umm now that I opened the PR I realise it makes no sense.

Issue #600, opened 2023-11-16
[BUG] LangChain OpenSearchVectorSearch broken with 2.4.1 w/AOSS


mirodrr:
### What is the bug?
_A clear and concise description of the bug._
When using LangChain OpenSearchVectorSearch with OpenSearchServerless and doing:
```
docsearch = OpenSearchVectorSearch(index_name=ELASTICSEARCH_INDEX,
                                       embedding_function=embeddings,
                                       opensearch_url=f'https://{serverless_collection_host}:443',
                                       **opensearch_serverless_args)
docsearch.add_texts(all_text_split, metadatas=metadatas)
```

OR

```
docsearch = OpenSearchVectorSearch.from_texts(
                all_text_split,
                embeddings,
                metadatas=metadatas,
                opensearch_url=f'https://{serverless_collection_host}:443',
                bulk_size=12000,
                index_name=ELASTICSEARCH_INDEX,
                **opensearch_serverless_args
            )
```

I get the error
```
'status': 400, 'error': {'type': 'illegal_argument_exception', 'reason': 'Document ID is not supported in create/index operation request'}
```

This same code works fine on 2.3.2.

### How can one reproduce the bug?
_Steps to reproduce the behavior._
Create a "OpenSearchVectorSearch" in LangChain when using opensearch-py 2.4.0, and try to upload documents to it

### What is the expected behavior?
_A clear and concise description of what you expected to happen._
Documents upload successfully
### What is your host/environment?
_Operating system, version._
Python Lambda image. Specifically `public.ecr.aws/lambda/python:3.9`

### Do you have any screenshots?
_If applicable, add screenshots to help explain your problem._
No
### Do you have any additional context?
_Add any other context about the problem._
No

VachaShah:
I tried adding id="1" to https://github.com/opensearch-project/opensearch-py/blob/main/test_opensearchpy/test_server/test_helpers/test_data.py#L39 which is used in tests in this repo. The tests that use this method fail with the following error:

```
args = (<opensearchpy.client.indices.IndicesClient object at 0x7f3432d28e20>,)
kwargs = {'body': {'mappings': {'properties': {'author': {'properties': {'name': {...}}}, 'authored_date': {'type': 'date'}, 'c...path_hierarchy', 'type': 'custom'}}}, 'number_of_replicas': 0, 'number_of_shards': 1}}, 'id': '1', 'index': 'flat-git'}
[...]

PR #547, opened 2023-10-20
Added support for AWS Sigv4 for UrlLib3.

dblock:
### Description

- Adds SigV4 support for Urllib3HttpConnection.
- Deprecates `AWSV4SignerAuth` in favor of `RequestsAWSV4SignerAuth`
- Uses `Urllib3AWSV4SignerAuth` in the documentation since that transport is faster.

#### AOS

```
export SERVICE=es
export ENDPOINT=https://xyz.us-west-2.aoss.amazonaws.com
poetry run python sigv4/urllib3-hello.py 

INFO:Found credentials in environment variables.
INFO:GET https://search-dblock-test-opensearch-27-xyz.us-west-2.es.amazonaws.com:443/ [status:200 request:0.768s]
opensearch: 2.7.0
INFO:PUT https://search-dblock-test-opensearch-27-xyz.us-west-2.es.amazonaws.com:443/movies [status:200 request:2.298s]
INFO:PUT https://search-dblock-test-opensearch-27-xyz.us-west-2.es.amazonaws.com:443/movies/_doc/1 [status:201 request:0.500s]
INFO:POST https://search-dblock-test-opensearch-27-xyz.us-west-2.es.amazonaws.com:443/_search [status:200 request:0.213s]
{'director': 'Bennett Miller', 'title': 'Moneyball', 'year': 2011}
INFO:DELETE https://search-dblock-test-opensearch-27-xyz.us-west-2.es.amazonaws.com:443/movies/_doc/1 [status:200 request:0.108s]
INFO:DELETE https://search-dblock-test-opensearch-27-xyz.us-west-2.es.amazonaws.com:443/movies [status:200 request:0.223s]
```

#### AOSS

```
export SERVICE=aoss
export ENDPOINT=https://xyz.us-west-2.aoss.amazonaws.com
poetry run python sigv4/urllib3-hello.py 
```

```
INFO:Found credentials in environment variables.
INFO:PUT https://xyz.us-west-2.aoss.amazonaws.com:443/movies [status:200 request:4.173s]
INFO:PUT https://xyz.us-west-2.aoss.amazonaws.com:443/movies/_doc/1 [status:201 request:3.117s]
INFO:POST https://xyz.us-west-2.aoss.amazonaws.com:443/_search [status:200 request:4.916s]
{'director': 'Bennett Miller', 'title': 'Moneyball', 'year': 2011}
[...]

PR #603, opened 2023-11-17
Fix: re-add service to fix integration with LangChain.

dblock:
### Description

Re-add `.service`, which is [used by LangChain](https://github.com/langchain-ai/langchain/blob/b4312aac5c0567088353178fb70fdb356b372e12/libs/langchain/langchain/vectorstores/opensearch_vector_search.py#L83) to figure out whether we're talking to AOSS. This broke in https://github.com/opensearch-project/opensearch-py/pull/547.

### Issues Resolved

Closes https://github.com/opensearch-project/opensearch-py/issues/600

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Removed
 ### Fixed
 - Fix `TypeError` on `parallel_bulk` ([#601](https://github.com/opensearch-project/opensearch-py/pull/601))
+- Fix Amazon OpenSearch Serverless integration with LangChain ([#603](https://github.com/opensearch-project/opensearch-py/pull/603))
 ### Security
 
 ## [2.4.1]
--- a/opensearchpy/helpers/signer.py
+++ b/opensearchpy/helpers/signer.py
@@ -78,6 +78,7 @@ class RequestsAWSV4SignerAuth(requests.auth.AuthBase):
 
     def __init__(self, credentials, region, service: str = "es") -> None:  # type: ignore
         self.signer = AWSV4Signer(credentials, region, service)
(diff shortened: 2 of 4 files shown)


navneet1v:
@dblock its not necessary related to AOSS. Even in AOS, if user is asking for OpenSearch service to generate the ids for the data client should not send _id in the bulk request.

dblock:
> @dblock its not necessary related to AOSS. Even in AOS/OSS, if user is asking for OpenSearch service to generate the ids for the data, os client should not send _id in the bulk request.

The code is pretty explicit :) 

```
def _is_aoss_enabled(http_auth: Any) -> bool:
    """Check if the service is http_auth is set as `aoss`."""
    if (
        http_auth is not None
        and hasattr(http_auth, "service")
        and http_auth.service == "aoss"
    ):
        return True
    return False
```

harshavamsi (approved):

navneet1v:
> > @dblock its not necessary related to AOSS. Even in AOS/OSS, if user is asking for OpenSearch service to generate the ids for the data, os client should not send _id in the bulk request.
> 
> The code is pretty explicit :)
> 
> ```python
> def _is_aoss_enabled(http_auth: Any) -> bool:
>     """Check if the service is http_auth is set as `aoss`."""
>     if (
>         http_auth is not None
>         and hasattr(http_auth, "service")
>         and http_auth.service == "aoss"
>     ):
>         return True
>     return False
> ```

This code is explicit, because AOSS doesn't support _id. But given that client is generic and AOSS support _id for search collection. So if we make this change then search collections will break.

@dblock 

This is the reason why I am saying the expectation from opensearch-py is if user has not provided _id, client should not add of its own whether it is AOS, OSS or AOSS.

dblock:
> This is the reason why I am saying the expectation from opensearch-py is if user has not provided _id, client should not add of its own whether it is AOS, OSS or AOSS.

Agreed. It does not. Am I missing anything?

PR #604, opened 2023-11-17
Fixes a suspicious setattr that should be taking an Any

### LINKED ISSUE #602: Incompatible types in assignment for document fields
### What is the bug?
types on fields are not inferred correctly when setting from instance methods

a contrived example:
```python
class MyDocument(Document):
    myVal = Double()

    def save(self, **kwargs):
        # set default value for myVal
        if not self.myVal:
            self.myVal = 1
        return super().save(**kwargs)
```

running mypy would yield:
```
error: Incompatible types in assignment (expression has type "int", variable has type "Double")  [assignment]
```

no type error

dblock:
### Description

Fixes a suspicious `setattr` that should be taking an `Any`.

The `Double` supports `__setattr__` assignment with a float, but I am not sure how to express that in the actual `Double` class.

### Issues Resolved

Related to #602.

### DIFF
--- a/test_opensearchpy/test_helpers/test_document.py
+++ b/test_opensearchpy/test_helpers/test_document.py
@@ -32,7 +32,7 @@ import ipaddress
 import pickle
 from datetime import datetime
 from hashlib import sha256
-from typing import Any
+from typing import Any, Union
 
 from pytest import raises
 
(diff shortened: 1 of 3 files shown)


test_opensearchpy/test_helpers/test_document.py
```diff
@@ -648,3 +648,28 @@ class MySubDocWithNested(MyDoc):
         },
         "title": {"type": "keyword"},
     }
+
+
+def test_save_double(mock_client: Any) -> None:
+    class MyDocumentWithDouble(MyDoc):
+        a_double: Union[float, field.Double] = field.Double()
```
harshavamsi, line 655: why is this only float and not `field.Float`?

harshavamsi (approved):
I think this looks good. Wondering how the float field is coming in.

test_opensearchpy/test_helpers/test_document.py
dblock: One is a primitive type and the other one is a field.

PR #616, opened 2023-11-25
Added fix for key error because of missing 'hits' key.

Djcarrillo6:
### Description
Adds graceful/defensive code for handling case of missing "hits" key in scan response.

### Issues Resolved
Resolves issue [#313](https://github.com/opensearch-project/opensearch-py/issues/313)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Removed
 - Removed unnecessary `# -*- coding: utf-8 -*-` headers from .py files ([#615](https://github.com/opensearch-project/opensearch-py/pull/615))
 ### Fixed
+- Fix KeyError when scroll return no hits ([#616](https://github.com/opensearch-project/opensearch-py/pull/616))
 ### Security
 
 ## [2.4.2]
--- a/opensearchpy/helpers/actions.py
+++ b/opensearchpy/helpers/actions.py
@@ -586,14 +586,17 @@ def scan(
     scroll_id = resp.get("_scroll_id")
 
     try:
-        while scroll_id and resp["hits"]["hits"]:
-            for hit in resp["hits"]["hits"]:
+        while scroll_id and resp.get("hits", {}).get("hits"):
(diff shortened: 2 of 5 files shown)


CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Removed
 - Removed unnecessary `# -*- coding: utf-8 -*-` headers from .py files ([#615](https://github.com/opensearch-project/opensearch-py/pull/615))
 ### Fixed
+- Fixed key error bug which threw an exception when scroll is used and search results are empty because the hits key is not within hits dictionary ([#616](https://github.com/opensearch-project/opensearch-py/pull/616))
```
dblock, line 15: Let's get more specific and less verbose. How about something like "Fix `KeyError` when `scroll` return no `hits`"? We don't need to explain why in CHANGELOG.

dblock (commented):
AFAIK tests should go next to other scroll tests, so in `test_actions.py`? 

Btw, tests aren't very well organized, my rule of thumb is that for every file like `actions.py` there should be one `test_actions.py` and if actions is a kitchen sink of many things, i can be split up into a folder called `actions` with files such as `scroll.py`, and tests organized similarly. So whatever gets us closer to that is A-OK.

opensearchpy/helpers/actions.py
```diff
@@ -586,14 +586,17 @@ def scan(
     scroll_id = resp.get("_scroll_id")
 
     try:
-        while scroll_id and resp["hits"]["hits"]:
-            for hit in resp["hits"]["hits"]:
+        while scroll_id and resp.get("hits", {}).get("hits"):
```
saimedhi, line 589: Hi @Djcarrillo6, I've reviewed the code. Thanks for your contribution! Could you please make similar adjustments in the async helpers as well?

Djcarrillo6: @saimedhi absolutely!

dblock: You should be able to run `nox -rs generate` and it will make all those changes for you I believe.

dblock (approved):

PR #615, opened 2023-11-24
remove unnecessary utf-8 header in .py files

samuelorji:
### Description
Remove Unnecessary `# -*- coding: utf-8 -*-`  headers from .py files

### Issues Resolved
https://github.com/opensearch-project/opensearch-py/issues/613

By submitting this pull request, I confirm that my contribution is made under the terms of the Apache 2.0 license.
For more information on following Developer Certificate of Origin and signing off your commits, please check [here](https://github.com/opensearch-project/OpenSearch/blob/main/CONTRIBUTING.md#developer-certificate-of-origin).

PR #617, opened 2023-11-28
Remove unnecessary utf-8 header in license_headers.py (follow-up to #615)

### LINKED ISSUE #613: [BUG] Missing or unnecessary utf-8 header in .py files
### What is the bug?

In #557 we added UTF-8 headers, but lost them in https://github.com/opensearch-project/opensearch-py/commit/bcfef113c43f74c41dec9c43ba54d6e00cc17c90 again. Do we need them?

Either have them or don't have them, and add a linter that ensures headers are present/not present in all files.

jayaddison:
### Description
This is a small followup to @samuelorji's cleanup in #615.  It applies the same cleanup (removal of a redundant/default `coding` declaration of `utf-8` in source files) to the `license_headers.py` check script itself.

This pull request also removes that line from the list of `LINES_TO_KEEP` - lines of source code that aren't checked by the license check script - since it shouldn't appear anymore.

### Issues Resolved
#613

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,7 +10,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Changed
 ### Deprecated
 ### Removed
-- Removed unnecessary `# -*- coding: utf-8 -*-` headers from .py files ([#615](https://github.com/opensearch-project/opensearch-py/pull/615))
+- Removed unnecessary `# -*- coding: utf-8 -*-` headers from .py files ([#615](https://github.com/opensearch-project/opensearch-py/pull/615), [#617](https://github.com/opensearch-project/opensearch-py/pull/617))
 ### Fixed
 ### Security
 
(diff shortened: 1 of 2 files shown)


dblock (approved):
Good catch, thank you.

saimedhi (approved):

saimedhi:
@jayaddison, I think changelog entry is not needed.

dblock:
I opened two more issues related to API generator, #619 and #618, if you have any interest @jayaddison :)

jayaddison:
Thanks @saimedhi - I had pushed a commit with a changelog entry a few moments before seeing that you added a label to skip it.  It was merged - I hope that's OK?

@dblock those issues do look interesting, thanks; I cannot offer any sense of when I might work on them yet, though.

Something that I notice about the `utils/generate_api.py` script is that it could be difficult to easily confirm what version (`git` commit ID) of the `OpenSearch.openapi.json` spec file was used to produce the output.  Has including a version/commit identifier, or referencing the spec file using a `git submodule` been considered?

dblock:
I would largely prefer a submodule for the exact reason you describe. Feel free to open a feature request or just do it (TM).

PR #627, opened 2023-12-11
[draft] Use a git submodule to include the OpenSearch.openapi.json file

### LINKED ISSUE #626: [PROPOSAL] Use a git submodule to include the OpenSearch.openapi.json file
### What are you proposing?
The [`generate_api.py` utility script](https://github.com/opensearch-project/opensearch-py/blob/db61b5943bdc0582b8f310581470d13e87f7413f/utils/generate_api.py) currently [retrieves](https://github.com/opensearch-project/opensearch-py/blob/db61b5943bdc0582b8f310581470d13e87f7413f/utils/generate_api.py#L432-L434) the latest version of the OpenSearch API spec based on the `main` branch of the [`opensearch-api-specification.git` repository](https://github.com/opensearch-project/opensearch-api-specification/).

In the case of updating the contents of the development tip in `opensearch-py` (this repo), that's what we want, but there are a few scenarios that are challenging:

  * Developing `opensearch-py` against a branched version of the specifications (perhaps rare / unexpected).
  * Regenerating a historic version of the client code to debug some past issue (also perhaps rare).

The proposed solution to this is to use a `git` submodule to reference a _specific version_ of `opensearch-api-specification.git`.  The submodule contents would be cloned locally as part of the existing `nox -rs generate` command.

### What users have asked for this feature?
As far as I know, I'm the only requestor so far.

### What problems are you trying to solve?
In addition to the two items mentioned above (development against a branched spec, debugging of past issues), I like to increase the confidence I have that a software release can be rebuilt again in the future.  Specifying dependency versions tends to helps a lot for that - but it does have a tradeoff which is that those versions need to be managed.

### What is the developer experience going to be?
This effect of this change is on developers of the `opensearch-py` library itself.

  * The good: the process to autogenerate client code from the spec would be unchanged (`nox -rs generate`).
  * The not so great: the version of `opensearch-api-specification.git` embedded as a submodule would need to be adjusted when the client is being upgraded to match a new spec.

#### Are there any security considerations? 
The changes here remove one HTTP(S) request, and replace it with a local file read to a `git` submodule checkout.

The content we receive shouldn't change in either case, and the access level is gated -- and should be equivalent -- whether a `git` clone of the submodule occurs or based on the HTTP request, because the repository is public.

One additional subprocess is created within the `nox` step -- a call to the local `git` binary on the host system.

#### Are there any breaking changes to the API
No.

The `opensearch-py` developer experience when generating client code from the OpenSearch spec should be unchanged.

When an updated version of the OpenSearch API specs become available, an additional manual step is required to update the submodule to the desired version; this update is then commited using `git commit -m "..."` or similar as with any other change entry.

#### Are there breaking changes to the User Experience?
Developers could unexpectedly find that when they run the `nox -rs generate` script, they do not see recent updates from the spec repository without explicitly updating to include those.

This could improve some of the development flexibility for the client code in this repository; however there is no evidence of branching in the spec repository, so it is unclear whether this is valuable in practice.

The proposal could make it easier to revisit historic versions of the client and regenerate code as it existed at the time.

Generally this proposal intends to make it easier to replicate a precise build of the `opensearch-py` client reliably.

I have some prototype code available in a branch, and I'll offer this in a 'draft' status until there's consensus on whether the feature is worthwhile.

### Any remaining open questions?
The process for updating the submodule version is undefined.

### Additional context
This proposal branches off from some [discussion on pull request #617](https://github.com/opensearch-project/opensearch-py/pull/617#issuecomment-1830627719).

jayaddison:
### Description
Instead of using an HTTP(S) request to retrieve the latest `main` OpenSearch.openapi.json spec from `opensearch-api-specification.git`, reference it as a `git` submodule and read the contents from file.

### Issues Resolved
Resolves #626.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "specifications/opensearch-api-specification"]
+	path = specifications/opensearch-api-specification
+	url = https://github.com/opensearch-project/opensearch-api-specification.git/
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added a log collection guide ([#579](https://github.com/opensearch-project/opensearch-py/pull/579))
 - Added GHA release ([#614](https://github.com/opensearch-project/opensearch-py/pull/614))
 ### Changed
+- Read the `OpenSearch.openapi.json` API specification file from a `git` submodule checkout instead of making an HTTP request. ([#627](https://github.com/opensearch-project/opensearch-py/pull/627))
 ### Deprecated
 ### Removed
 - Removed unnecessary `# -*- coding: utf-8 -*-` headers from .py files ([#615](https://github.com/opensearch-project/opensearch-py/pull/615), [#617](https://github.com/opensearch-project/opensearch-py/pull/617))
--- a/dev-requirements.txt
+++ b/dev-requirements.txt
@@ -1,4 +1,3 @@
-requests>=2, <3
 pytest
 pytest-cov
 coverage
--- a/noxfile.py
+++ b/noxfile.py
@@ -118,6 +118,8 @@ def docs(session: Any) -> None:
 
 @nox.session()  # type: ignore
 def generate(session: Any) -> None:
+    session.run("git", "submodule", "init", external=True)
+    session.run("git", "submodule", "update", external=True)
     session.install("-rdev-requirements.txt")
     session.run("python", "utils/generate_api.py")
     format(session)
--- a/utils/generate_api.py
+++ b/utils/generate_api.py
@@ -41,7 +41,6 @@ from typing import Any, Dict
 
 import black
 import deepmerge
-import requests
 import unasync
 import urllib3
 from click.testing import CliRunner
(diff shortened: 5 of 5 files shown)


dblock:
I like it! Let's document this in .md file(s), CHANGELOG, etc.

saimedhi:
Hello @jayaddison, could you please address the failing tests and implement the requested changes? Let's work towards getting this PR merged.

jayaddison:
Thanks for the ping @saimedhi.  I'd paused here because I'm not certain that my `git` submodule proposal (#626) is gaining traction.  Even so, it seems reasonable to keep up-to-date, and I see that there are changes going on related to API-change-detection.

Should I merge `main` and resolve conflicts too?

dblock:
@Xtansia had some opinions about this direction, what do you recommend we do with this PR?

Xtansia:
> @Xtansia had some opinions about this direction, what do you recommend we do with this PR?

@dblock @jayaddison Since we don't yet have a versioning & release process in place for the specs, I'm not completely against this approach for now as it's easy to change in the future. It's important to keep in mind the current proposal to swap the spec repo away from Smithy and at the same time a decent chance the final one-file "complete" spec will not be checked into Git and instead be released in some other manner.

dblock:
I think @Xtansia is saying that we won't have a submodule reference if/when we switch to OpenAPI. Let's see what happens with https://github.com/opensearch-project/opensearch-api-specification/issues/189 before we decide on this?

saimedhi:
> I think @Xtansia is saying that we won't have a submodule reference if/when we switch to OpenAPI. Let's see what happens with [opensearch-project/opensearch-api-specification#189](https://github.com/opensearch-project/opensearch-api-specification/issues/189) before we decide on this?

@dblock, @Xtansia: Now that [opensearch-project/opensearch-api-specification#189](https://github.com/opensearch-project/opensearch-api-specification/issues/189) has been resolved, could you please share your thoughts on this PR?

dblock:
The spec is no longer committed to git but published as a release, so let's close it.

PR #633, opened 2023-12-14
Updated the get_policy API in the index_management plugin 

### LINKED ISSUE #629: [FEATURE] Add support for returning all index policy objects
Currently there is a nice integration for the [index management plugin](https://opensearch-project.github.io/opensearch-py/api-ref/plugins/index_management_plugin.html#opensearchpy.plugins.index_management.IndexManagementClient) supporting APIs like [get_policy](https://opensearch-project.github.io/opensearch-py/api-ref/plugins/index_management_plugin.html#opensearchpy.plugins.index_management.IndexManagementClient.get_policy). However, that `get_policy()` API has "policy" as required positional argument. 

As a result, there is no way to request all the policies that live in a cluster.

Make "policy" an optional argument and return all policies whenever it is not given (or maybe also when policy is set to `*`, in case we wanna go down that path).
Alternatively, add a new API `get_policies`, which will behave like the alerting plugin's [get_alerts](https://opensearch-project.github.io/opensearch-py/api-ref/plugins/alerting_plugin.html#opensearchpy.plugins.alerting.AlertingClient.get_alerts). 

In any case, the call should behave like running `GET _plugins/_ism/policies` on the cluster, thus returning:
```
{
  "policies": [
      { },
      ...
      { }
  ],
  "total_policies": 7
}
```

Can't find a way that could fetch all policies of a cluster.

N/A

saimedhi:
### Description
Updated the `get_policy` API in the index_management plugin to allow the policy_id argument as optional

### Issues Resolved
Closes #629 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added a log collection guide ([#579](https://github.com/opensearch-project/opensearch-py/pull/579))
 - Added GHA release ([#614](https://github.com/opensearch-project/opensearch-py/pull/614))
 ### Changed
+- Updated the `get_policy` API in the index_management plugin to allow the policy_id argument as optional ([#633](https://github.com/opensearch-project/opensearch-py/pull/633))
 ### Deprecated
 ### Removed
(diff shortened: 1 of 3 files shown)


VachaShah (commented):
@saimedhi Is it not possible to generate the plugin endpoints instead of manually fixing them?

dblock (approved):

saimedhi:
> @saimedhi Is it not possible to generate the plugin endpoints instead of manually fixing them?

Yes @VachaShah, we need specifications and a few modifications to the generator to generate plugins in the plugins folder.

VachaShah (approved):

PR #639, opened 2023-12-21
Reusable async client

### LINKED ISSUE #637: [FEATURE] Re-create `aiohttp.ClientSession` after using `close()` on `AsyncOpensearch`
I find it pretty annoying to handle the `close()` method of `AsyncOpensearch`. Basically, once closed, the client becomes unusable because every attempt to connect to the Opensearch cluster will yield the error `opensearchpy.exceptions.ConnectionError: ConnectionError(Session is closed) caused by: RuntimeError(Session is closed)`.

This happens because `close()` effectively closes the `aiohttp.ClientSession` without removing the attribute from `AsyncHttpSession`. A simple solution would be to write `self.session = None` right after https://github.com/opensearch-project/opensearch-py/blob/e323ab2435a26ffd21104a2c765f9711b8614c3a/opensearchpy/connection/http_async.py#L279.

Re-creating the `AsyncOpensearch` instance, which bring unecessary overhead?

How to create a `RuntimeError`:

```python
import asyncio
from opensearchpy import AsyncOpenSearch
loop = asyncio.get_event_loop()
opensearch = AsyncOpenSearch(...)
async def get_health():
    print(await opensearch.cluster.health())

loop.run_until_complete(get_health())         # Ok
loop.run_until_complete(opensearch.close())   # Closing
loop.run_until_complete(get_health())         # RuntimeError
```

odelmarcelle:
### Description
Clears out the `session` attribute of `AsyncHttpConnection` and `AIOHttpConnection` when calling `close()`, which allow re-using a previously closed `AsyncOpensearch` instance.

### Issues Resolved
Closes #637

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,12 +8,12 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added a log collection guide ([#579](https://github.com/opensearch-project/opensearch-py/pull/579))
 - Added GHA release ([#614](https://github.com/opensearch-project/opensearch-py/pull/614))
 ### Changed
-- `AsyncOpensearch` can now be re-used after calling `close()` ([#639](https://github.com/opensearch-project/opensearch-py/pull/639))
 ### Deprecated
 ### Removed
 - Removed unnecessary `# -*- coding: utf-8 -*-` headers from .py files ([#615](https://github.com/opensearch-project/opensearch-py/pull/615), [#617](https://github.com/opensearch-project/opensearch-py/pull/617))
 ### Fixed
 - Fix KeyError when scroll return no hits ([#616](https://github.com/opensearch-project/opensearch-py/pull/616))
+- Fix reuse of `AsyncOpenSearch` after calling `close` ([#639](https://github.com/opensearch-project/opensearch-py/pull/639))
 ### Security
 ### Dependencies
 - Bumps `pytest-asyncio` from <=0.21.1 to <=0.23.2
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added a log collection guide ([#579](https://github.com/opensearch-project/opensearch-py/pull/579))
 - Added GHA release ([#614](https://github.com/opensearch-project/opensearch-py/pull/614))
 ### Changed
+- `AsyncOpensearch` can now be re-used after calling `close()` ([#635](https://github.com/opensearch-project/opensearch-py/pull/635))
 ### Deprecated
 ### Removed
 - Removed unnecessary `# -*- coding: utf-8 -*-` headers from .py files ([#615](https://github.com/opensearch-project/opensearch-py/pull/615), [#617](https://github.com/opensearch-project/opensearch-py/pull/617))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added a log collection guide ([#579](https://github.com/opensearch-project/opensearch-py/pull/579))
 - Added GHA release ([#614](https://github.com/opensearch-project/opensearch-py/pull/614))
 ### Changed
-- `AsyncOpensearch` can now be re-used after calling `close()` ([#635](https://github.com/opensearch-project/opensearch-py/pull/635))
+- `AsyncOpensearch` can now be re-used after calling `close()` ([#639](https://github.com/opensearch-project/opensearch-py/pull/639))
 ### Deprecated
 ### Removed
 - Removed unnecessary `# -*- coding: utf-8 -*-` headers from .py files ([#615](https://github.com/opensearch-project/opensearch-py/pull/615), [#617](https://github.com/opensearch-project/opensearch-py/pull/617))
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -8,6 +8,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added a log collection guide ([#579](https://github.com/opensearch-project/opensearch-py/pull/579))
 - Added GHA release ([#614](https://github.com/opensearch-project/opensearch-py/pull/614))
 ### Changed
+- `AsyncOpensearch` can now be re-used after calling `close()` ([#639](https://github.com/opensearch-project/opensearch-py/pull/639))
```
dblock, line 11: Nit: I think this is a bug, so I would move this to a fix rather than something changed, "Fix reuse of `AsyncOpenSearch` after calling `close`." Note the spelling of OpenSearch ;)

dblock (changes requested):
LGTM! Thanks. 

Double checking that the synchronous client doesn't have this issue?

Does the client support `with` (too lazy to look) :))

odelmarcelle:
Turns out the issue affects clients using `Urllib3HttpConnection` which discard the pool, but `RequestsHttpConnection` works fine with my local testing.

I discovered this by adding test for the synchronous client, which apparently uses `Urllib3HttpConnection`. I'm not too sure where to put a test that would use `RequestsHttpConnection`.

saimedhi:
> Turns out the issue affects clients using `Urllib3HttpConnection` which discard the pool, but `RequestsHttpConnection` works fine with my local testing.
> 
> I discovered this by adding test for the synchronous client, which apparently uses `Urllib3HttpConnection`. I'm not too sure where to put a test that would use `RequestsHttpConnection`.


Hello @odelmarcelle , please put your test [here](https://github.com/opensearch-project/opensearch-py/blob/main/test_opensearchpy/test_client/test_requests.py) for RequestsHttpConnection. And also DCO is failing. Please fix it. Thank you.

odelmarcelle:
Resolved the DCO.

@saimedhi AFAIK, the test won't work there as there's not connection established to an opensearch instance. It should be somewhere under /test_server/, right? I didn't manage to work it out by myself.

One last comment though: Why are there two async connection classes defined? I'm not too sure to understand why `AsyncHttpConnection` and `AIOHttpConnection` co-exist. Especially as the former is the one recommended in the guides while most tests are based on the latter.

opensearchpy/connection/http_urllib3.py
```diff
@@ -214,9 +214,14 @@ def __init__(
         if pool_maxsize and isinstance(pool_maxsize, int):
             kw["maxsize"] = pool_maxsize
 
-        self.pool = pool_class(
-            self.hostname, port=self.port, timeout=self.timeout, **kw
-        )
+        def urllib3_pool_factory() -> Any:
```
dblock, line 217: Why is this not at the class level? Is it because of `**kw`?

test_opensearchpy/test_async/test_server/test_clients.py
```diff
@@ -67,3 +67,14 @@ async def test_aiohttp_connection_works_without_yarl(
...
         resp = await async_client.info(pretty=True)
         assert isinstance(resp, dict)
+
+
+class TestClose:
+    async def test_close_doesnt_break_client(self, async_client: Any) -> None:
+        await async_client.close()
+        await async_client.cluster.health()
+
+        async with async_client:
```
dblock, line 77: Split into its own test.

test_opensearchpy/test_server/test_clients.py
```diff
@@ -49,3 +49,15 @@ def test_bulk_works_with_bytestring_body(self) -> None:
...
         self.assertEqual(1, len(response["items"]))
+
+
+class TestClose(OpenSearchTestCase):
+    def test_close_doesnt_break_client(self) -> None:
+        self.client.cluster.health()
+        self.client.close()
+        self.client.cluster.health()
+
+        with self.client as client:
```
dblock, line 60: Split into its own test, tests are cheap.

Maybe a range iteration 2 times is nicer? I wish we had the Ruby equivalent of `2.times do`. Now I'm really nitpicking :)

opensearchpy/connection/http_urllib3.py
odelmarcelle: Yes, do you having any better suggestion?

dblock: Use a `lambda`?

odelmarcelle: That triggers a warning in the linter because the lambda is not type annotated

dblock: You can suppress it with a comment?

odelmarcelle: done

dblock (approved):

dblock:
LMK if you prefer to merge as is, or if you think adding that `assert` is a good idea.

opensearchpy/connection/http_urllib3.py
```diff
@@ -214,9 +214,13 @@ def __init__(
...
             kw["maxsize"] = pool_maxsize
 
-        self.pool = pool_class(
+        self._urllib3_pool_factory = lambda: pool_class(
             self.hostname, port=self.port, timeout=self.timeout, **kw
         )
+        self._create_urllib3_pool()
+
+    def _create_urllib3_pool(self) -> None:
+        self.pool = self._urllib3_pool_factory()  # type: ignore
```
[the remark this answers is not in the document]
odelmarcelle: I'm not sure to see how that would help. `_urllib3_pool_factory` will never be `None`?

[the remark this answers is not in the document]
dblock: Yes, you're right, its creation is in the constructor.

PR #654, opened 2024-01-19
Use .pylintrc files

### LINKED ISSUE #645: [PROPOSAL] Replace lint_per_folder function in noxfile.py with .pylint files
It may be more standard to use a .pylint file to specify pylint overrides. Advantages could be fewer runs of pylint, more standard. Disadvantages could be that all of those exceptions are peppered throughout the code. 

Unless it's trivial this doesn't need to be part of this PR, but maybe there's a more generic solution where a .pylint file is placed into each folder with the configuration for that folder? This code could iterate over the folders and run it with the contents of that config file.

_Originally posted by @dblock in https://github.com/opensearch-project/opensearch-py/pull/643#discussion_r1446203242_

macohen:
### Description
switches pylint to use .pylintrc configs instead of python generated pylint commands

### Issues Resolved
#645 

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -25,7 +25,7 @@
 #  under the License.
 
 
-from typing import Any, List
+from typing import Any
 
 import nox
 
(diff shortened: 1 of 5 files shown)


noxfile.py
```diff
@@ -102,7 +102,11 @@ def lint(session: Any) -> None:
     session.run("black", "--check", *SOURCE_FILES)
     session.run("flake8", *SOURCE_FILES)
 
-    lint_per_folder(session)
+    pylint_overrides = ["opensearchpy/", "test_opensearchpy/"]
```
dblock, line 105: Instead of this find all .pylintrc files under first level subfolders and run those?

macohen:
The one thing I don't like about this solution is that each .pylintrc stands on its own. There's no way to inherit the config from a higher level folder so if max-line-length should be 240 across the whole project, it needs to be in each .pylintrc.

dblock (approved):
A lot better.

Issue #610, opened 2023-11-21
[FEATURE] Enable pylint defaults


dblock:
### Is your feature request related to a problem?

We enabled pylint in #590. Enable the remaining lints. Here are some top ones.

```
$ python -m pylint setup.py noxfile.py opensearchpy/ test_opensearchpy/ utils/ samples/ benchmarks/ docs/ | grep '(' | grep '-' | rev | cut -d '(' -f1 | rev | cut -d ')' -f1 | sort | uniq -c | sort -r

1304 missing-function-docstring
 540 missing-class-docstring
 484 protected-access
 477 duplicate-code
 223 no-member
 203 missing-module-docstring
 143 unused-argument
 101 consider-using-f-string
  87 super-with-arguments
  78 too-few-public-methods
  77 redefined-builtin
  69 line-too-long
  68 too-many-arguments
  66 redefined-outer-name
  48 import-outside-toplevel
  41 useless-object-inheritance
  35 unused-variable
  32 unexpected-keyword-arg
  25 raise-missing-from
  24 too-many-locals
  23 unnecessary-dunder-call
  20 wrong-import-position
  18 too-many-public-methods
  18 invalid-unary-operand-type
  18 attribute-defined-outside-init
  17 unspecified-encoding
  16 no-else-return
  16 invalid-overridden-method
  15 cyclic-import
  11 too-many-branches
  11 dangerous-default-value
  11 arguments-renamed
  10 pointless-statement
```
 
### What solution would you like?

Enable lints in setup.cfg one-by-one.

macohen:
So, to confirm, when you say one-by-one, it's this approach:
1. add a lint
2. run pylint
3. fix ALL the failures
4. go to 1 until out of lints or I just can't anymore. ;)

macohen:
Doing this incrementally, what do you think of breaking this down into an issue per lint? It becomes a much less daunting single task and can be picked up in parallel.

dblock:
@macohen Yes, I don't expect all the lints to be enabled at once, that's a massive change.

macohen:
Running utils/build_dists.py results in an error:

Installing collected packages: urllib3, six, idna, charset-normalizer, certifi, requests, python-dateutil, opensearch-py2
Successfully installed certifi-2023.11.17 charset-normalizer-3.3.2 idna-3.6 opensearch-py2-2.4.3 python-dateutil-2.8.2 requests-2.31.0 six-1.16.0 urllib3-1.26.18
[...]

PR #658, opened 2024-01-22
Assignment from no return

macohen:
### Description
* enables assignment-from-no-return
~* updated to pass this by marking Facet.get_value_filter(self, filter_value: Any) -> Any as an abc.abstractmethod~
~* implemented these missing methods in two subclasses to return None~
* wondering if we would be better off not enforcing this, but I'm leaning towards enforcement

### Issues Resolved
progress on #610

### DIFF
--- a/test_opensearchpy/run_tests.py
+++ b/test_opensearchpy/run_tests.py
@@ -94,9 +94,13 @@ def fetch_opensearch_repo() -> None:
         # if the run is interrupted from a previous run, it doesn't clean up, and the git add origin command
         # errors out; this allows the test to continue
         remote_origin_already_exists = 3
-        print(e)
-        if e.returncode != remote_origin_already_exists:
-            sys.exit(1)
+        if e.returncode == remote_origin_already_exists:
+            print(
+                "Consider setting TEST_OPENSEARCH_NOFETCH=true if you want to reuse the existing local OpenSearch repo"
+            )
+        else:
+            print(e)
+        sys.exit(1)
 
     # fetch the sha commit, version from info()
     print("Fetching opensearch repo...")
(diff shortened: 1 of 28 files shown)


saimedhi:
Few tests are failing. Please try to make them pass.

test_opensearchpy/run_tests.py
```diff
@@ -94,9 +94,13 @@ def fetch_opensearch_repo() -> None:
...
-        print(e)
-        if e.returncode != remote_origin_already_exists:
-            sys.exit(1)
+        if e.returncode == remote_origin_already_exists:
+            print(
+                "Consider setting TEST_OPENSEARCH_NOFETCH=true if you want to reuse the existing local OpenSearch repo"
+            )
+        else:
+            print(e)
+        sys.exit(1)
```
saimedhi, line 103: should this `sys.exit(1)` be moved into else?

saimedhi (commented):
@macohen, could you kindly implement the requested changes and address the lint error? Once done, we can proceed with merging the code. Thank you!

test_opensearchpy/run_tests.py
macohen: My intent here was to make it clear why the exit was happening. I didn't want to change the previous behavior especially after I found the TEST_OPENSEARCH_NOFETCH variable.

saimedhi: @macohen, I apologize if I am causing any confusion, but I believe there's a change in the behavior for `e.returncode == remote_origin_already_exists`.

It seems that there shouldn't be a `sys.exit(1)` when `e.returncode == remote_origin_already_exists`. Please correct me if I'm mistaken.

macohen: Sorry for the poor explanation on my end. Prior to my change, when the git repository already had a remote origin added, the whole process exited. I was considering just ignoring that state, but then found the TEST_OPENSEARCH_NOFETCH variable which can be set when you want that state to be ignored. I figured it would be better to maintain the behavior of the script in that case and also tell the user that TEST_OPENSEARCH_NOFETCH could help if exiting was not desired. I hope that helps...

saimedhi (approved):

PR #660, opened 2024-01-24
Incorporated API generation into CI workflow and fixed 'generate' nox session

### LINKED ISSUE #618: [FEATURE] run nox -rs generate in CI

The generator may be modifying code when run that corresponds to a change currently being made. 

We want to make sure the generator runs clean as part of every PR. Run `nox -rs generate` in CI and fail if there are more changes to files.

saimedhi:
### Description
Incorporated API generation into CI workflow and fixed 'generate' nox session

### Issues Resolved
Closes #618 

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -18,6 +18,21 @@ jobs:
           python3.7 -m pip install nox
       - name: Lint the code
         run: nox -s lint
+  
+  generate:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Repository
+        uses: actions/checkout@v3
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.7
+      - name: Install dependencies
+        run: |
+          python3.7 -m pip install nox
+      - name: Run the api generator
+        run: nox -s generate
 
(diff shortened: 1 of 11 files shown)


saimedhi:
@VachaShah, can you please take a look at this PR.

.github/workflows/ci.yml
```diff
@@ -18,6 +18,21 @@ jobs:
...
+        uses: actions/checkout@v3
+      - name: Set up Python 3.7
+        uses: actions/setup-python@v4
+        with:
+          python-version: 3.7
+      - name: Install dependencies
+        run: |
+          python3.7 -m pip install nox
+      - name: Run the api generator
+        run: nox -s generate
```
dblock, line 35: I think we also want to block the PR if there were changes, don't we? Can be done in a separate PR.

dblock (approved):

.github/workflows/ci.yml
saimedhi: > I think we also want to block the PR if there were changes, don't we

- We will add a daily cron job to update code if there are specification changes. 

- PRs shouldn't be blocked as long as the generator CI test passes, even if they include changes, daily cron job handles it.

PR #677, opened 2024-02-20
Add NeuralQuery

philippemnoel:
Hi there!

This is a bit dated (a few months), but I figured I'd PR anyway in case it's valuable to the community. Back in August 2023, we forked this repository to add the NeuralQuery ability (combining full-text and vector search) as a precursor to what became https://github.com/paradedb/paradedb.

We've had the repo archived since then, and who knows perhaps this is of use to the community, so opening up here!

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+  "[python]": {
+    "editor.defaultFormatter": "ms-python.black-formatter"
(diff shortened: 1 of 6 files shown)


dblock:
Thanks. Are you interested in finishing the PR? We probably should discuss what this DSL looks like, add tests, etc.

Issue #678, opened 2024-02-21
[FEATURE]Add "Service Time" Measurement in opensearch-py


saimedhi:
### Is your feature request related to a problem?
Opensearch-py currently lacks service time measurement functionality, hindering performance analysis.

### What solution would you like?
I propose adding service time measurement functionality to opensearch-py, returning service time data conditionally upon user request.

### What alternatives have you considered?

- I haven't found a Pythonic solution for this yet. Any guidance in this regard would be greatly appreciated.
- I'm exploring various approaches and will update with draft PRs soon.

saimedhi:
Need to add support for async client

PR #679, opened 2024-02-21
`Service Time` measurement Trail_Approach_1

saimedhi:
### Description

- Service time is calculated within the perform_request function of the Connection class ( This draft uses RequestsHttpConnection) at the point where the server call is made.

- The calculated service time is added to the response or data in the connection. 

- I understand that this method is not ideal for returning service time and am exploring alternative options for improvement.

### Issues Resolved
Related to #678 

### DIFF
--- a/opensearchpy/connection/http_requests.py
+++ b/opensearchpy/connection/http_requests.py
@@ -163,7 +163,6 @@ class RequestsHttpConnection(Connection):
         body: Optional[bytes] = None,
         timeout: Optional[Union[int, float]] = None,
         allow_redirects: Optional[bool] = True,
-        calculate_service_time: Optional[bool] = False,
         ignore: Collection[int] = (),
         headers: Optional[Mapping[str, str]] = None,
     ) -> Any:
--- a/opensearchpy/connection/http_requests.py
+++ b/opensearchpy/connection/http_requests.py
@@ -166,6 +166,10 @@ class RequestsHttpConnection(Connection):
         ignore: Collection[int] = (),
         headers: Optional[Mapping[str, str]] = None,
     ) -> Any:
+        
+        calculate_service_time=False
+        if "calculate_service_time" in self.kwargs:
+            calculate_service_time=self.kwargs["calculate_service_time"]
         url = self.base_url + url
         headers = headers or {}
         if params:
--- a/opensearchpy/connection/http_requests.py
+++ b/opensearchpy/connection/http_requests.py
@@ -88,6 +88,7 @@ class RequestsHttpConnection(Connection):
         pool_maxsize: Any = None,
         **kwargs: Any
     ) -> None:
+        self.kwargs=kwargs
         if not REQUESTS_AVAILABLE:
             raise ImproperlyConfigured(
                 "Please install requests to use RequestsHttpConnection."
--- a/opensearchpy/connection/http_requests.py
+++ b/opensearchpy/connection/http_requests.py
@@ -188,7 +188,9 @@ class RequestsHttpConnection(Connection):
         }
         send_kwargs.update(settings)
         try:
+            request_start = time.perf_counter()
             response = self.session.send(prepared_request, **send_kwargs)
+            request_duration=time.perf_counter()-request_start
             duration = time.time() - start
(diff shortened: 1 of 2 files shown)


saimedhi:
Sample usage
```
host = 'localhost'
port = 9200
auth = ('admin', 'admin') # For testing only. Don't store credentials in code.

client = OpenSearch(
    hosts = [{'host': host, 'port': port}],
    http_auth = auth,
    use_ssl = True,
    verify_certs = False,
    connection_class = RequestsHttpConnection,
    calculate_service_time = True,
)

info = client.info()
print("info         ", info)

```
info          {'name': 'c3b344c0372b', 'cluster_name': 'docker-cluster', 'cluster_uuid': 'f6p2z4YRQvGW8nknOPgPeg', 'version': {'distribution': 'opensearch', 'number': '2.11.1', 'build_type': 'tar', 'build_hash': '6b1986e964d440be9137eba1413015c31c5a7752', 'build_date': '2023-11-29T21:45:35.524809067Z', 'build_snapshot': False, 'lucene_version': '9.7.0', 'minimum_wire_compatibility_version': '7.10.0', 'minimum_index_compatibility_version': '7.0.0'}, 'tagline': 'The OpenSearch Project: https://opensearch.org/', 'x-service-time': '0.04581618309020996'}

opensearchpy/connection/http_requests.py
```diff
@@ -207,7 +213,11 @@ def perform_request(  # type: ignore
             if isinstance(e, requests.Timeout):
                 raise ConnectionTimeout("TIMEOUT", str(e), e)
             raise ConnectionError("N/A", str(e), e)
-
+        
+        # Add the service time to the raw_data
+        if calculate_service_time:
+            raw_data = raw_data.rstrip()[:-1] + f', "__client": {{"Service_Time": "{duration}"}}' + '}'
```
Xtansia, line 219: After a quick look, I think rather than attempting to modify the JSON string like this, it'd make more sense to just return the value from this function. ie. changing `return response.status_code, response.headers, raw_data` to `return response.status_code, response.headers, raw_data, duration` or creating some 'RequestMetrics` class that contains the duration (and potential additional future metrics) which is returned. Then in the transport class inserting that into the returned/deserialized object as `__metrics__` or something like that. 

Adding it to the return signature of the connection class gives a clear and uniform way of returning this information across all connection implementations.

Some one with more Python prowess will hopefully be able to give a better idea of what's actually idiomatic.

VachaShah: +1 for this.

opensearchpy/connection/http_requests.py
```diff
@@ -187,13 +191,15 @@ def perform_request(  # type: ignore
             "allow_redirects": allow_redirects,
         }
         send_kwargs.update(settings)
+        start = time.time()
```
VachaShah, line 194: I think the start at line 179 is fine. Any reason otherwise?

opensearchpy/connection/http_requests.py
```diff
@@ -187,13 +191,15 @@ def perform_request(  # type: ignore
...
         try:
             response = self.session.send(prepared_request, **send_kwargs)
             duration = time.time() - start
             raw_data = response.content.decode("utf-8", "surrogatepass")
         except reraise_exceptions:
             raise
         except Exception as e:
+            duration = time.time() - start
```
VachaShah, line 202: I don't see this variable being used anywhere since `ConnectionError` is raised after.

opensearchpy/connection/http_requests.py
saimedhi: Hello @VachaShah, I am adding start timer just before sending request. I think we don't need  timer to start before prepared_request right?

dblock:
This is a good start and shows what we're trying to achieve.

What do you think about an idea where we first expose "events", then add "metrics"? Googling I found https://stackoverflow.com/questions/443885/python-callbacks-delegates-what-is-common and https://pypi.org/project/Events/ that could be a start. The client could expose events such as "request started", "finished", "errored", etc., and then another class called `Metrics` or `Measurements` could collect time information using this event system.

PR #684, opened 2024-03-04
Enhance generator to update changelog only if generated code differs from existing

saimedhi:
### Description
Enhance generator to update changelog only if generated code differs from existing code. 

### Issues Resolved
https://github.com/opensearch-project/opensearch-py/pull/680#issuecomment-1966865315

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Added GHA release ([#614](https://github.com/opensearch-project/opensearch-py/pull/614))
 - Incorporated API generation into CI workflow and fixed 'generate' nox session ([#660](https://github.com/opensearch-project/opensearch-py/pull/660))
 - Added an automated api update bot for opensearch-py ([#664](https://github.com/opensearch-project/opensearch-py/pull/664)) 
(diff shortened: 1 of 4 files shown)


dblock (approved):
I'll merge, but this seems a little complicated. 

Would it make things simpler to do the `rm -rf`, generate the code into `opensearchpy`, then look at `git status` to see what has changed?

saimedhi:
> I'll merge, but this seems a little complicated.
> 
> Would it make things simpler to do the `rm -rf`, generate the code into `opensearchpy`, then look at `git status` to see what has changed?

I'll give that approach a try. Just want to make sure files like .pycache won't cause any issues.

dblock:
> > I'll merge, but this seems a little complicated.
> > Would it make things simpler to do the `rm -rf`, generate the code into `opensearchpy`, then look at `git status` to see what has changed?
> 
> I'll give that approach a try. Just want to make sure files like .pycache won't cause any issues.

Anything generated should be in `.gitignore`.

dblock:
> I'll give that approach a try.

Btw, I think it's not important _what_ has changed. The workflow that runs this check could generate the code, then check whether anything has changed, and only if something has generate the changelog. Then try opening the PR which would do nothing if there were no changes.

PR #689, opened 2024-03-06
Introduced service time metrics to OpenSearch-Py client.

### LINKED ISSUE #678: [FEATURE]Add "Service Time" Measurement in opensearch-py

saimedhi:
### Description
Introduced service time metrics to OpenSearch-Py client.

### Issues Resolved
#678 

##### Guide and Tests need to be added for this change.

### DIFF
--- a/opensearchpy/connection/http_requests.py
+++ b/opensearchpy/connection/http_requests.py
@@ -36,6 +36,8 @@ try:
 except ImportError:
     REQUESTS_AVAILABLE = False
 
+from opensearchpy.metrics import Metrics
+
 from ..compat import reraise_exceptions, string_types, urlencode
 from ..exceptions import (
     ConnectionError,
@@ -86,8 +88,10 @@ class RequestsHttpConnection(Connection):
         http_compress: Any = None,
         opaque_id: Any = None,
         pool_maxsize: Any = None,
+        metrics: Optional[Metrics] = None,
         **kwargs: Any
     ) -> None:
+        self.metrics = metrics
         if not REQUESTS_AVAILABLE:
             raise ImproperlyConfigured(
                 "Please install requests to use RequestsHttpConnection."
@@ -188,7 +192,11 @@ class RequestsHttpConnection(Connection):
         }
         send_kwargs.update(settings)
         try:
+            if self.metrics is not None:
+                self.metrics.request_start()
             response = self.session.send(prepared_request, **send_kwargs)
+            if self.metrics is not None:
+                self.metrics.request_end()
             duration = time.time() - start
             raw_data = response.content.decode("utf-8", "surrogatepass")
         except reraise_exceptions:
@@ -244,7 +252,15 @@ class RequestsHttpConnection(Connection):
             duration,
         )
 
-        return response.status_code, response.headers, raw_data
+        if self.metrics is None:
+            return response.status_code, response.headers, raw_data
+        else:
+            return (
+                response.status_code,
+                response.headers,
+                raw_data,
+                self.metrics.service_time,
+            )
 
     @property
     def headers(self) -> Any:  # type: ignore
--- a/opensearchpy/metrics.py
+++ b/opensearchpy/metrics.py
@@ -1,0 +1,65 @@
+# SPDX-License-Identifier: Apache-2.0
+#
+# The OpenSearch Contributors require contributions made to
+# this file be licensed under the Apache-2.0 license or a
+# compatible open source license.
+#
+# Modifications Copyright OpenSearch Contributors. See
+# GitHub history for details.
+
+import time
+from abc import ABC, abstractmethod
+
+from events import Events
+
+
+class Metrics(ABC):
+    @abstractmethod
+    def request_start(self) -> None:
+        pass
+
+    @abstractmethod
+    def request_end(self) -> None:
+        pass
+
+    @property
+    @abstractmethod
+    def start_time(self) -> float:
+        pass
+
+    @property
+    @abstractmethod
+    def service_time(self) -> float:
+        pass
+
+
+class MetricsEvents(Metrics):
+    @property
+    def start_time(self) -> float:
+        return self._start_time
+
+    @property
+    def service_time(self) -> float:
+        return self._service_time
+
+    def __init__(self) -> None:
+        self.events = Events()
+        self._start_time = 0.0
+        self._service_time = 0.0
+
+        # Subscribe to the request_start and request_end events
+        self.events.request_start += self._on_request_start
+        self.events.request_end += self._on_request_end
+
+    def request_start(self) -> None:
+        self.events.request_start()
+
+    def _on_request_start(self) -> None:
+        self._start_time = time.perf_counter()
+
+    def request_end(self) -> None:
+        self.events.request_end()
+
+    def _on_request_end(self) -> None:
+        self._end_time = time.perf_counter()
+        self._service_time = self._end_time - self._start_time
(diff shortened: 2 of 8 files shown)


saimedhi:
I'm unsure about my changes to async connections. I've tested them, but I'm uncertain if I've covered all cases. Let's start by applying these changes for requests and urllib3 connections first. For async connections, I'll need to review whether a similar approach to sync connections is suitable or if we should utilize something like aiohttptraceconfig.

opensearchpy/connection/http_requests.py
```diff
@@ -187,9 +189,17 @@ def perform_request(  # type: ignore
             "allow_redirects": allow_redirects,
         }
         send_kwargs.update(settings)
+
+        calculate_service_time = False
```
dblock, line 193: This seems very specific. I propose to add `metrics` parameter that would default to `None` or an instance of `Metrics`. I'd like to be able to write 

```py
client = Client(metrics = Metrics())
```

Then the `if` is not necessary.

opensearchpy/connection/http_requests.py
```diff
@@ -187,9 +189,17 @@ def perform_request(  # type: ignore
...
+        if "calculate_service_time" in self.kwargs:
+            calculate_service_time = self.kwargs["calculate_service_time"]
+
+        time_metrics = TimeMetrics()
```
dblock, line 197: Create an instance variable for the client called `self.metrics`

opensearchpy/connection/http_requests.py
```diff
@@ -187,9 +189,17 @@ def perform_request(  # type: ignore
...
+
         try:
+            time_metrics.events.server_request_start()
```
dblock, line 200: You shouldn't b reaching inside `.events`, it should be something like `self.metrics.request_start()`.

opensearchpy/metrics.py
```diff
@@ -0,0 +1,33 @@
...
+# Modifications Copyright OpenSearch Contributors. See
+# GitHub history for details.
+
+
+import time
+
+from events import Events
+
+
+class TimeMetrics:
```
dblock, line 16: Create an interface and a class that implements it.

https://www.geeksforgeeks.org/abstract-classes-in-python/

```py
class Metrics(ABC)
   @abstractmethod
   def request_start()
```

```py
class MetricsEvents(Metrics):
   def request_start(self):
      self.start_time ...
```

dblock (changes requested):
Good start. The `if calculate_service_time:` is the anti-pattern we need to fix here and make it more object-oriented. Define a generic `Metrics` interface with methods such as `request_start`, and a specific implementation for it using `Events`, e.g. `EventsMetrics`. See below.

opensearchpy/connection/http_requests.py
```diff
@@ -187,9 +190,14 @@ def perform_request(  # type: ignore
             "allow_redirects": allow_redirects,
         }
         send_kwargs.update(settings)
+        self.is_metrics_none = False
```
dblock, line 193: This looks like an anti-pattern. Either `self.metrics` is set, or it's not. Default to not.

saimedhi: @dblock, If it's not used, then [self.metrics.request_start()](https://github.com/opensearch-project/opensearch-py/pull/689/files#diff-1e38589c63abdaf216eafb88d5419b81729acb2c5ccc71d5b4af7a52da105e47R277) and [self.metrics.request_end()](https://github.com/opensearch-project/opensearch-py/pull/689/files#diff-1e38589c63abdaf216eafb88d5419b81729acb2c5ccc71d5b4af7a52da105e47R281) can't be calculated. Are you suggesting we keep [start = time.time()](https://github.com/opensearch-project/opensearch-py/pull/689/files#diff-1e38589c63abdaf216eafb88d5419b81729acb2c5ccc71d5b4af7a52da105e47L245) and [duration = time.time() - start](https://github.com/opensearch-project/opensearch-py/pull/689/files#diff-1e38589c63abdaf216eafb88d5419b81729acb2c5ccc71d5b4af7a52da105e47L274) untouched, and only calculate and return service_time when self.metrics is not None?

saimedhi: Made the changes as requested. Working on adding tests.

opensearchpy/connection/http_requests.py
```diff
@@ -188,7 +192,11 @@ def perform_request(  # type: ignore
         }
         send_kwargs.update(settings)
         try:
+            if self.metrics is not None:
+                self.metrics.request_start()
             response = self.session.send(prepared_request, **send_kwargs)
+            if self.metrics is not None:
```
dblock, line 198: If there's an exception this is never called. Use `finally` to always `request_end`. Add a test.

opensearchpy/connection/http_requests.py
```diff
@@ -244,7 +252,15 @@ def perform_request(  # type: ignore
...
-        return response.status_code, response.headers, raw_data
+        if self.metrics is None:
+            return response.status_code, response.headers, raw_data
+        else:
+            return (
+                response.status_code,
+                response.headers,
+                raw_data,
+                self.metrics.service_time,
+            )
```
dblock, line 263: I don't think you should be returning any of the metrics, the caller can look into `.metrics.service_time` if they need to.

opensearchpy/metrics.py
```diff
@@ -0,0 +1,65 @@
...
+    def start_time(self) -> float:
+        pass
+
+    @property
+    @abstractmethod
+    def service_time(self) -> float:
+        pass
+
+
+class MetricsEvents(Metrics):
```
dblock, line 36: Make a new folder/namespace called `metrics`, and separate these classes into individual files.

opensearchpy/metrics.py
```diff
@@ -0,0 +1,65 @@
...
+    def start_time(self) -> float:
+        return self._start_time
+
+    @property
+    def service_time(self) -> float:
+        return self._service_time
+
+    def __init__(self) -> None:
+        self.events = Events()
+        self._start_time = 0.0
```
dblock, line 47: Any reason to not expose these? I would just do `self.start_time`.

These should begin with `None`.

opensearchpy/metrics.py
```diff
@@ -0,0 +1,65 @@
...
+
+        # Subscribe to the request_start and request_end events
+        self.events.request_start += self._on_request_start
+        self.events.request_end += self._on_request_end
+
+    def request_start(self) -> None:
+        self.events.request_start()
+
+    def _on_request_start(self) -> None:
+        self._start_time = time.perf_counter()
```
dblock, line 58: Re-assign `self._end_time = None` here. If metrics is reused, you don't want to get wrong time. Add a test.

opensearchpy/transport.py
```diff
@@ -457,6 +478,10 @@ def perform_request(
                     data = self.deserializer.loads(
                         data, headers_response.get("content-type")
                     )
+
+                if self.metrics:
```
dblock, line 482: Why is this needed? You can always look into metrics, can't you?

dblock (commented):
Better. Needs tests and some more changes.

Issue #698, opened 2024-03-19
[BUG] AsyncOpenSearch Authorization Error When Using AWSV4SignerAsyncAuth


vcastane:
### What is the bug?
Generating an AsyncOpenSearch client with AWSV4SignerAsyncAuth/AsyncHttpConnection causes authorization errors despite valid credentials using an assumed session

### How can one reproduce the bug?
Authorization differences can be seen when comparing a sync opensearch client to an async opensearch client:

Sync Opensearch:
```
# Sync

import boto3
from opensearchpy import OpenSearch, AWSV4SignerAuth, RequestsHttpConnection

arn_to_assume = "<your-arn-here>"
region = "<region>"
role_session_name = "<role-session-name>"
opensearch_endpoint = "<my-dev-endpoint.amazonaws.com>"
port = int(443)  # it's probably 443

base_session = boto3.Session()
sts = base_session.client('sts')

sts_response = sts.assume_role(RoleArn=arn_to_assume, RoleSessionName=role_session_name)

assumed_session = boto3.Session(
    aws_access_key_id=sts_response['Credentials']['AccessKeyId'],
    aws_secret_access_key=sts_response['Credentials']['SecretAccessKey'],
    aws_session_token=sts_response['Credentials']['SessionToken']
)

credentials = assumed_session.get_credentials()

auth = AWSV4SignerAuth(credentials, region)
client = OpenSearch(
    hosts = [{'host': opensearch_endpoint, 'port': port}],
    http_auth = auth,
    use_ssl = True,
    verify_certs = True,
    connection_class = RequestsHttpConnection
)

client.ping()
```
This gets the expected `200` response and a `True`

When attempting to recreate the same behavior with the AsyncOpenSearch client as below:
```
# Async

import boto3
from opensearchpy import AsyncOpenSearch, AWSV4SignerAsyncAuth, AsyncHttpConnection

arn_to_assume = "<your-arn-here>"
region = "<region>"
role_session_name = "<role-session-name>"
opensearch_endpoint = "<my-dev-endpoint.amazonaws.com>"
port = int(443)  # it's probably 443

base_session = boto3.Session()
sts = base_session.client('sts')

sts_response = sts.assume_role(RoleArn=arn_to_assume, RoleSessionName=role_session_name)

assumed_session = boto3.Session(
    aws_access_key_id=sts_response['Credentials']['AccessKeyId'],
    aws_secret_access_key=sts_response['Credentials']['SecretAccessKey'],
[...]

PR #699, opened 2024-03-20
FIX: Remove HEAD -> GET request swap workaround

vcastane:
### Description
Remove a workaround replacing HEAD requests with GET requests for async requests.
This workaround is no longer needed after the underlying bug was fixed in https://github.com/aio-libs/aiohttp/pull/5012

### Issues Resolved
Partially resolves https://github.com/opensearch-project/opensearch-py/issues/698
The rest of the issue is user (my) error.

### DIFF
--- a/opensearchpy/_async/http_aiohttp.py
+++ b/opensearchpy/_async/http_aiohttp.py
@@ -246,12 +246,8 @@ class AIOHttpConnection(AsyncConnection):
         else:
             query_string = ""
 
-        # There is a bug in aiohttp that disables the re-use
-        # of the connection in the pool when method=HEAD.
-        # See: aio-libs/aiohttp#1769
         is_head = False
         if method == "HEAD":
-            method = "GET"
             is_head = True
 
         # Top-tier tip-toeing happening here. Basically
@@ -301,9 +297,9 @@ class AIOHttpConnection(AsyncConnection):
                 timeout=timeout,
                 fingerprint=self.ssl_assert_fingerprint,
             ) as response:
-                if is_head:  # We actually called 'GET' so throw away the data.
+                if is_head:
                     await response.release()
-                    raw_data = ""
+                    raw_data = ""  # HEAD method has no response body
                 else:
                     raw_data = await response.text()
                 duration = self.loop.time() - start
--- a/opensearchpy/connection/http_async.py
+++ b/opensearchpy/connection/http_async.py
@@ -167,12 +167,8 @@ class AsyncHttpConnection(AIOHttpConnection):
         else:
             query_string = ""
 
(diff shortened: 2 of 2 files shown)


opensearchpy/_async/http_aiohttp.py
```diff
@@ -301,9 +297,9 @@ async def perform_request(
                 timeout=timeout,
                 fingerprint=self.ssl_assert_fingerprint,
             ) as response:
-                if is_head:  # We actually called 'GET' so throw away the data.
+                if is_head:
                     await response.release()
-                    raw_data = ""
+                    raw_data = ""  # HEAD method has no response body
```
dblock, line 302: Should be `None`? There other other places where we assign things to `""` where in fact we mean none I believe, too.

opensearchpy/_async/http_aiohttp.py
```diff
@@ -246,12 +246,8 @@ async def perform_request(
         else:
             query_string = ""
 
-        # There is a bug in aiohttp that disables the re-use
-        # of the connection in the pool when method=HEAD.
-        # See: aio-libs/aiohttp#1769
         is_head = False
```
dblock, line 249: I believe this is no longer needed as we only reuse it once, maybe just check the verb where it's needed?

dblock (commented):
Do we need to lock aiohttp >= a certain version as part of the PR?

dblock:
Is there a way to write a test for this, too? Something that fails without this change?

saimedhi:
> Do we need to lock aiohttp >= a certain version as part of the PR?

- Bug resolved in aiohttp version [3.7.0](https://pypi.org/project/aiohttp/3.7.0/).
- opensearch-py compatibility: aiohttp>=3,<4.
- I suggest instead of locking aiohttp to >=3.7.0,<4, we should check the aiohttp version and conditionally remove the workaround for aiohttp >=3.7.0,<4.
- Detailed code review required to identify necessary code changes.

saimedhi:
> * I suggest instead of locking aiohttp to >=3.7.0,<4, we should check the aiohttp version and conditionally remove the workaround for aiohttp >=3.7.0.

Disregard the previous comment. To resolve this [issue](https://github.com/opensearch-project/opensearch-py/issues/715), we're planning to lock aiohttp to >=3.9.2,<4. Thus, simply removing the workaround should suffice."

saimedhi:
Hello @vcastane ,

Could you please take a look at finishing this PR when you have a moment? If you need any help, let us know.

Thanks a lot!

dblock:
Finished this in https://github.com/opensearch-project/opensearch-py/pull/794.

PR #716, opened 2024-04-03
Introduced service time metrics to OpenSearch-Py client.

### LINKED ISSUE #678: [FEATURE]Add "Service Time" Measurement in opensearch-py

saimedhi:
### Description
Introduced service time metrics to OpenSearch-Py client.

### Issues Resolved
#678 

- [x] Added tests
- [ ] Guide to be added 
- [ ] Service time metrics will be implemented for AsyncOpensearch in a subsequent PR


saimedhi:
The failing tests are not related to this PR but are caused by recent changes to the OpenSearch API Specification, which is now natively written in OpenAPI. The generator and developer guide will be fixed accordingly in an upcoming PR.

saimedhi:
@VachaShah please take a look

opensearchpy/connection/http_urllib3.py
```diff
@@ -115,8 +117,10 @@ def __init__(
         ssl_context: Any = None,
         http_compress: Any = None,
         opaque_id: Any = None,
+        metrics: Optional[Metrics] = None,
         **kwargs: Any
     ) -> None:
+        self.metrics = metrics
```
dblock, line 123: Does it make sense to do `self.metrics = Metrics()` or `self.metrics = MetricsNone()` that doesn't do anything to avoid having to `if self.metrics` all over the place?

dblock (dismissed):
Rebase? This LGTM.

opensearchpy/connection/http_urllib3.py
saimedhi: Hello @dblock, 

We can't use `self.metrics = Metrics()` because Metrics is an abstract class. However, we can create a subclass called MetricsNone that sets all values to None for service_time, start_time, and end_time. Then, we'll set default metrics to MetricsNone(). Does this approach align with your thoughts?

dblock: Yes. I think that would be better and future proof, no?

saimedhi:
Ready for review

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ## [Unreleased]
 ### Added
 - Added support for Python 3.12 ([#717](https://github.com/opensearch-project/opensearch-py/pull/717))
+- Introduced `service time` metrics to opensearch-py client ([#716](https://github.com/opensearch-project/opensearch-py/pull/716))
```
dblock, line 7: Remove "to opensearch-py client", this whole repo is the client.

"Added service time metrics" 

to be consistent with other changelog entries

opensearchpy/connection/http_requests.py
```diff
@@ -86,8 +91,10 @@ def __init__(
         http_compress: Any = None,
         opaque_id: Any = None,
         pool_maxsize: Any = None,
+        metrics: Any = MetricsNone(),
```
dblock, line 94: Instead of `Any` we know that `metrics` is a `Metrics` and is required. Double check that you cannot pass `None` (should produce a mypy error)

opensearchpy/connection/http_urllib3.py
```diff
@@ -115,8 +120,10 @@ def __init__(
         ssl_context: Any = None,
         http_compress: Any = None,
         opaque_id: Any = None,
+        metrics: Any = MetricsNone(),
```
dblock, line 123: `metrics: Metrics`

opensearchpy/transport.py
```diff
@@ -91,6 +93,7 @@ class Transport(object):
     last_sniff: float
     sniff_timeout: Optional[float]
     host_info_callback: Any
+    metrics: Any
```
dblock, line 96: metrics: Metrics

opensearchpy/transport.py
```diff
@@ -112,6 +115,7 @@ def __init__(
         retry_on_status: Collection[int] = (502, 503, 504),
         retry_on_timeout: bool = False,
         send_get_body_as: str = "GET",
+        metrics: Any = MetricsNone(),
```
dblock, line 118: metrics: Metrics

dblock (changes requested):
Type `Metrics` when passed around, `metrics: Metrics`.

dblock (approved):

Issue #628, opened 2023-12-12
[FEATURE] Add support for urllib3 > v2


Bialogs:
### What is the bug?

When using `opensearch-py` v2.3.2 and greater with `boto3` provided by AWS Lambda , the following error occurs:

`cannot import name 'DEFAULT_CIPHERS' from 'urllib3.util.ssl_'`

### How can one reproduce the bug?

Dockerfile:

```Dockerfile
FROM public.ecr.aws/lambda/python:3.9
RUN pip3 install opensearch-py==2.3.1 --target ${LAMBDA_TASK_ROOT}
COPY /app/ ${LAMBDA_TASK_ROOT}/
CMD ["app.function"]
```

app.py:

```python
import boto3
import opensearchpy

def function(event, ctx)
    print("hello")
```

### What is the expected behavior?

Imports are successful and code runs.

### What is your host/environment?

`public.ecr.aws/lambda/python:3.9`

### Do you have any screenshots?

N/A

### Do you have any additional context?

Downgrade to 2.3.1 successfully executes.

Bialogs:
Yes from my test it failed with v2.3.2 through v2.4.2.

In terms of the runtime I'm not installing any other packages other than `opensearch-py` at the specified versions. I'm noticing that the commits for v2.3.2 include changes to the `urllib3` version. Perhaps that introduced this?

Bialogs:
Examining the base image, `public.ecr.aws/lambda/python:3.9` the following `urllib3` is bundled. I assume because it is integrated with the other software packages that the base image contains.

From Docker Scout (`docker scout sbom public.ecr.aws/lambda/python:3.9`) 
````
    {
      "type": "pypi",
      "name": "urllib3",
      "version": "1.26.18",
      "purl": "pkg:pypi/urllib3@1.26.18",
      "author": "Andrey Petrov",
      "licenses": [
        "MIT"
      ],
      "locations": [
        {
          "path": "/var/runtime/urllib3-1.26.18.dist-info/METADATA",
          "ordinal": 7,
          "digest": "sha256:4dd1a48d6392cbd0f03fdefc227609f129f9648bfc83a06f8435f8536eb1efbe",
          "diff_id": "sha256:8a810533d2d18a464c851d93a9aabf4ddbcdf6c171bef202acaeec45d29e9779"
        },
        {
          "path": "/var/runtime/urllib3-1.26.18.dist-info/RECORD",
          "ordinal": 7,
[...]

Issue #695, opened 2024-03-13
New opensearch py release


VijayanB:
We would like to update [ForceMerge()](https://github.com/opensearch-project/opensearch-benchmark/blob/main/osbenchmark/worker_coordinator/runner.py#L674) Runner from OpenSearch Benchmark with latest api https://github.com/opensearch-project/opensearch-py/pull/660 . This will help workloads like vector search to use force merge as async operation since force merge takes hours to complete for large dataset.

I created a workaround in local fork https://github.com/VijayanB/opensearch-benchmark/commit/b73d3a9de2b4dbc2d3a5bc071c1d47167ae5aec6 

This is blocker for vector search benchmark. It will be great if we could get this change from upstream instead of cherry-picking.

Issue #561, opened 2023-10-31
[FEATURE] Release v. Next


dblock:
### Is your feature request related to a problem?

Coming from https://github.com/opensearch-project/opensearch-py/issues/560, users are getting confused by newer docs.

### What solution would you like?

Release v. Next.

### What alternatives have you considered?

We should probably add some disclaimers to the README that the doc is newer than the last released version.

saimedhi:
@dblock, before the release should we fix #558

dblock:
No strong opinions about that one.

dblock:
https://github.com/opensearch-project/opensearch-py/issues/585

saimedhi:
opensearch-py 2.4.0 is yet to be released. And also please create a branch 2.4 (It helps in patch releases)

dblock:
> opensearch-py 2.4.0 is yet to be released. And also please create a branch 2.4 (It helps in patch releases)

Meaning a 2.4 at the v2.4.0 tag? I am following https://github.com/opensearch-project/opensearch-py/blob/main/RELEASING.md#releasing, any changes to that?

dblock:
We've released 2.4.1, please use that (or newer).

PR #719, opened 2024-04-16
Add support for urllib3 v2

iherasymenko:
### Description

Allows using urllib3 v2 if Python version is >= 3.10. The same technique is applied in https://github.com/boto/botocore/pull/3141/files.

### Issues Resolved

https://github.com/opensearch-project/opensearch-py/issues/628

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 ### Added
+- Added support for urllib3 2+ in Python 3.10+ ([#719](https://github.com/opensearch-project/opensearch-py/pull/719))
 - Added support for Python 3.12 ([#717](https://github.com/opensearch-project/opensearch-py/pull/717))
 - Added service time metrics ([#716](https://github.com/opensearch-project/opensearch-py/pull/716))
 - Added `search_pipeline` APIs and `notifications` plugin APIs ([#724](https://github.com/opensearch-project/opensearch-py/pull/724))
--- a/setup.py
+++ b/setup.py
@@ -54,7 +54,8 @@ packages = [
     if package == MODULE_DIR or package.startswith(MODULE_DIR + ".")
 ]
 install_requires = [
-    "urllib3>=1.26.18, <2",
+    'urllib3>=1.26.18,<1.27 ; python_version < "3.10"',
+    'urllib3>=1.26.18,!=2.2.0,<3 ; python_version >= "3.10"',
     "requests>=2.4.0, <3.0.0",
     "six",
     "python-dateutil",


saimedhi:
Hello @iherasymenko, thank you for your contribution. Could you kindly include a few pertinent tests, please?

dblock:
Maybe we can introduce an integration test that uses each version of the lib?

setup.py
```diff
@@ -54,7 +54,8 @@
     if package == MODULE_DIR or package.startswith(MODULE_DIR + ".")
 ]
 install_requires = [
-    "urllib3>=1.26.18, <2",
+    'urllib3>=1.26.18,<1.27 ; python_version < "3.10"',
```
SamStephens, line 57: Why `<1.27` rather than `<2`? Surely we want those who cannot use urllib3 2.x to at least get updates to 1.x?

setup.py
```diff
@@ -54,7 +54,8 @@
...
+    'urllib3>=1.26.18,!=2.2.0,<3 ; python_version >= "3.10"',
```
SamStephens, line 58: Why the specific exclusion of 2.2.0?

james-certn:
Naive and hardline thinking: why preserve backward urllib3 compatibility in opensearchpy at all? If someone is tied to an old version of urllib3, they can use an older version of opensearchpy.

I get the desire for flexibility and supporting the most users possible, giving the latest to all users - but at some point the overhead isn't worth it, surely.

---

That aside, how would one best test this PR's changes in opensearchpy?

dblock:
> Naive and hardline thinking: why preserve backward urllib3 compatibility in opensearchpy at all? If someone is tied to an old version of urllib3, they can use an older version of opensearchpy.
> 
> I get the desire for flexibility and supporting the most users possible, giving the latest to all users - but at some point the overhead isn't worth it, surely.

That's all there is to it, flexibility. Generally if something is possible we tend not to break it to reduce downstream pain. If there's a good reason not to, or it's too hard to maintain, then we can drop it.
 
> That aside, how would one best test this PR's changes in opensearchpy?

Since we're explicitly saying that we support multiple incompatible dependencies I think we need to introduce some way of doing integration tests where one vs. another version of the library is loaded.

If this is hard to do, I would be open to merging this as is. Let's get your questions answered on the compatible versions, I think those make a lot of sense.

setup.py
iherasymenko: This is mirrors what botocore did. 2.2.0 has some breaking changes.

https://github.com/boto/botocore/pull/3141

https://github.com/boto/botocore/issues/3138#issuecomment-1986927748

setup.py
iherasymenko: This also mirrors what `botocore` did. I believe the original rationale was that 1.27 may introduce some breaking changes.

iherasymenko:
@james-certn The `opensearch-py` library is frequently (if not always) used together with the AWS Python SDK (https://github.com/boto/botocore/). Preferably, this library stays compatible with the Python 3.9 Lambda deployments, i.e. we don't pin the version of urllib3 to >2.x (as the urllib3 version there is pinned to 1.x) and, at the same time, we don't leave the version of urllib3 unpinned as it caused https://github.com/opensearch-project/opensearch-py/issues/628. 

@dblock @saimedhi I'm thinking about adding a separate Integration Tests job that uses the `public.ecr.aws/lambda/python:3.9` container so that we can make sure in a Python 3.9 environment, the right version gets picked up and this version does not conflict with the bundled `boto3`/`botocore` AWS Python SDK. Would this be a reasonable testing approach?

dblock:
> @dblock @saimedhi I'm thinking about adding a separate Integration Tests job that uses the `public.ecr.aws/lambda/python:3.9` container so that we can make sure in a Python 3.9 environment, the right version gets picked up and this version does not conflict with the bundled `boto3`/`botocore` AWS Python SDK. Would this be a reasonable testing approach?

I would start simpler and definitely not attempt to actually make AWS calls (we want to do that, but setting up infrastructure that runs permanently is currently difficult, we need credits, etc.). Given a certain configuration (e.g. in a Pipfile), you end up with a specific version of the dependent library, hard-coded in a Python test.

james-certn:
I'm not sure that I've seen tests in projects for this kind of requirements configuration. Not that I'm against the idea, but it does seem (subjectively) irregular. Maybe I just don't get out enough. :)

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 ### Added
+- Added support for urllib3 2+ in Python 3.10+
```
dblock, line 6: Add the PR number here like the other lines, please?

dblock (dismissed):
I am good with merging this (please update CHANGELOG, see below) and open an issue for tests if you aren't going to attempt it.

CHANGELOG.md
iherasymenko: Updated.

iherasymenko:
@dblock Thanks. I updated the PR. I'll try to allocate some time to work on the tests this week.

dblock (approved):
Let's merge since this is the same as in botocore.

james-certn:
Sorry for the Github noise, but when would a release around this be planned? Right now, the 2.5.0 versioned package as is does block a subset of users from leveraging the latest and greatest.

A ballpark of a day, a week, a month - that's all I'm asking for. ;) Thank you!

saimedhi:
> Sorry for the Github noise, but when would a release around this be planned? Right now, the 2.5.0 versioned package as is does block a subset of users from leveraging the latest and greatest.
> 
> A ballpark of a day, a week, a month - that's all I'm asking for. ;) Thank you!

Hello @james-certn, could you please create an issue requesting a 2.6.0 release? Similar to issues #561 and #695. Thank you!

PR #731, opened 2024-04-22
Enhanced the generator to omit generating default parameter values when already present in param description.

saimedhi:
### Description

- Enhanced the generator to omit generating default parameter values when they are already specified in the parameter description.

### Issues Resolved
 Closes https://github.com/opensearch-project/opensearch-py/pull/727#discussion_r1573170642

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,6 +13,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Fixed
 - Updated code generator to use native OpenAPI specification ([#721](https://github.com/opensearch-project/opensearch-py/pull/721))
 ### Updated APIs
+- Updated opensearch-py APIs to reflect [opensearch-api-specification@61f8a38](https://github.com/opensearch-project/opensearch-api-specification/commit/61f8a382c6abc7d9bb0504a097b52323809095be)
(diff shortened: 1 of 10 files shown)


dblock:
Shouldn't the spec not include the default in the description if it has an `x-default...` field?

saimedhi:
> Shouldn't the spec not include the default in the description instead if it has an `x-default...` field (because they can effectively disagree).

- Got it! So, if an x-default value is provided, the default value shouldn't be included in the param description in spec. @nhtruong Can we have this change in spec? 

-  In spec, there are two ways to specify default values for parameters. I need to consider both cases during generation. I missed example 1 case. I will make necessary changes. Thanks. 

Example 1:
```
reindex::query.wait_for_active_shards:
  in: query
  name: wait_for_active_shards
  description: |-
    The number of shard copies that must be active before proceeding with the operation.
    Set to `all` or any positive integer up to the total number of shards in the index (`number_of_replicas + 1`).
  schema:
    $ref: '#/components/schemas/_common:WaitForActiveShards'
  style: form
  x-default: '1'
```

Example 2:
```
reindex::query.wait_for_completion:
  in: query
  name: wait_for_completion
  description: If `true`, the request blocks until the operation is complete.
  schema:
    type: boolean
    default: true       
  style: form

```

dblock:
@saimedhi Open an issue in opensearch-api-specification to move any "default is ..." to `schema` or `x-default` and possibly add a linter. Also don't let @nhtruong beat you to picking up and implementing it ;)

PR #774, opened 2024-07-14
Update guides-bulk example

keithrozario:
Fixed examples to make them working python code. (missing commas and quotes.

### Description
Example code in the docs were missing crucial commas and double quotes. It meant the code wouldn't run in Python. 

While it is a relatively small error that most Python developers will notice it, newcomers might struggle to find the error.

Hope this pull requests makes the example code easier to understand.

### Issues Resolved

### DIFF
--- a/guides/bulk.md
+++ b/guides/bulk.md
@@ -37,11 +37,11 @@ The client can also serialize an array of data into bulk-delimited JSON for you.
 
 ```python
 data = [
-    { "index": { "_index": "index-2022-06-08", "_id": 1 }}
-    { "name": "foo"} 
-    { "index": { "_index": "index-2022-06-09", "_id": 2 }}
-    { "name": "bar"}
-    { "index": { "_index": "index-2022-06-10", "_id": 3 }}
+    { "index": { "_index": "index-2022-06-08", "_id": 1 }},
+    { "name": "foo"}, 
+    { "index": { "_index": "index-2022-06-09", "_id": 2 }},
+    { "name": "bar"},
+    { "index": { "_index": "index-2022-06-10", "_id": 3 }},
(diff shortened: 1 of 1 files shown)


dblock:
Thank you!

I assume examples in https://github.com/opensearch-project/opensearch-py/tree/main/samples/bulk do actually run, but if you have a minute please do double-check.

Would love to find a way where the guides are tested, at least for syntax correctness.

dblock (approved):

dblock:
Link checker is failing, if you have a second to fix that it'd be amazing!

keithrozario:
Thanks for merging this.

1. I looked through the samples before submitting, and at least they don't have this same issue. But not sure about other issues :)
2. The linkchecker seems a bit odd though, the link in questions seems to work fine. Perhaps its an issue with the runner while the check was running:

image

keithrozario:
"Would love to find a way where the guides are tested, at least for syntax correctness."

One thing we could do (and we do for other repos as well) is to move all the code to separate .py files instead of inline in the docs. From there we can lint the .py files to ensure syntax correctness. If it's something the project is looking into, I can perhaps help with that as well.

dblock:
@keithrozario I think it could be helpful, however I also think people appreciate guides with code snippets in them. We already have working samples. Dunno, maybe it's not worth overthinking.

Another idea is to go the route of https://github.com/opensearch-project/documentation-website/issues/7700 where we generate samples and documentation from spec (https://github.com/opensearch-project/documentation-website/issues/7700), we have these end-to-end tests that run against a real OpenSearch in https://github.com/opensearch-project/opensearch-api-specification/tree/main/tests, they could be turned into working samples and docs mechanically.

PR #777, opened 2024-07-16
Skip elements with titles.

dblock:
### Description

A workaround for the change in https://github.com/opensearch-project/opensearch-api-specification/pull/416 to avoid adding `node_id_or_metric`. With https://github.com/opensearch-project/opensearch-api-specification/pull/434 the only API tests that need to be skipped are those that `catch` an error (missing index parameters). These fail as expected on the client side.


saimedhi:
@dblock, Please take a look at the failing tests. Thank you.

dblock:
The failure is unrelated caused by https://github.com/opensearch-project/OpenSearch/commit/29a3e2c980764f305f0ed5e858878bc7bb3dbe64. The `put_alias` method doesn't actually require index or alias, which can be specified in the body. 

What's the right way to fix this @saimedhi?

saimedhi:
> The failure is unrelated caused by [opensearch-project/OpenSearch@29a3e2c](https://github.com/opensearch-project/OpenSearch/commit/29a3e2c980764f305f0ed5e858878bc7bb3dbe64). The `put_alias` method doesn't actually require index or alias, which can be specified in the body. The tests added in that PR do that.
> 
> What's the right way to fix this @saimedhi? Where do I remove the fact that these parameters are required and can default to `None`?

Hi @dblock, for path params they can be changed [here](https://github.com/opensearch-project/opensearch-py/blob/main/utils/generate_api.py#L386)

dblock:
https://github.com/opensearch-project/opensearch-py/pull/778

saimedhi (approved):

opensearchpy/client/indices.py
```diff
@@ -861,23 +864,23 @@ def get_field_mapping(
     )
     def put_alias(
         self,
-        index: Any,
-        name: Any,
         body: Any = None,
+        index: Any = None,
+        name: Any = None,
```
Tenzer, line 869: FYI: This is a breaking change for anybody that were calling the method without naming the arguments.

dblock: Looks like a miss :(

Care to open a bug, or if you have time fix the code generator?

Tenzer: The number of users affected by this is probably limited, but perhaps it's worth calling it out in the release notes?

Tenzer: I have made a bug for it here: https://github.com/opensearch-project/opensearch-py/issues/803.

PR #780, opened 2024-07-18
Allow the use of percent string in Bool.__and__ method

### LINKED ISSUE #779: [BUG] `TypeError` error when using a `.filter(...)` on a `Search` object.
### What is the bug?
Get a `TypeError` error when using a `.filter("terms", ...)` on a `Search` object.

Having a str instead of an int for `Bool.filter` or `Bool.must` attributes when reach the `Bool.__and__` method.
This case can happen when using string based percent (eg: `"50%"`) instead of pure integer (eg: `50`) for `minimum_should_match`.

Accept `50`, `"50"` or `"50%"` as valid values.

A [TODO statement](https://github.com/opensearch-project/opensearch-py/blob/main/opensearchpy/helpers/query.py#L222) is present in the code for this specific case.
This TODO statement date back at least from the [opensearch-dsl 1.0](https://github.com/opensearch-project/opensearch-dsl-py/blob/1.0/opensearch_dsl/query.py#L224).

Godefroy-Amaury:
### Description
Allow the use of `50`, `"50"` or `"50%"` as valid value for `qx._min_should_match`/`min_should_match` in the `Bool.__and__` method.

Previously, if for any reason the `qx._min_should_match` was a string, a TypeError error was raise by the `if len(qx.should) <= min_should_match:`.

Still raise a ValueError if the given value is an invalid integer representation.

### Issues Resolved
Closes #779

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Removed
 ### Fixed
 - Fixed Search helper to ensure proper retention of the _collapse attribute in chained operations. ([#771](https://github.com/opensearch-project/opensearch-py/pull/771))
+- Fixed the use of `minimum_should_match` with `Bool` to allow the use of string-based value (percent string, combination). ([#780](https://github.com/opensearch-project/opensearch-py/pull/780))
 ### Updated APIs
 - Updated opensearch-py APIs to reflect [opensearch-api-specification@0b033a9](https://github.com/opensearch-project/opensearch-api-specification/commit/0b033a92cac4cb20ec3fb51350c139afc753b089)
 - Updated opensearch-py APIs to reflect [opensearch-api-specification@d5ca873](https://github.com/opensearch-project/opensearch-api-specification/commit/d5ca873d20ff54be16ec48e7bd629cda7c4a6332)
--- a/opensearchpy/helpers/query.py
+++ b/opensearchpy/helpers/query.py
@@ -219,10 +219,12 @@ class Bool(Query):
                 del q._params["minimum_should_match"]
 
             for qx in (self, other):
-                # TODO: percentages will fail here
                 min_should_match = qx._min_should_match
                 # all subqueries are required
-                if len(qx.should) <= min_should_match:
+                if (
+                    isinstance(min_should_match, int)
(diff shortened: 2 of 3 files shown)


dblock:
Thanks! Add a test and a CHANGELOG entry, please?

It would be amazing if you could please look at https://github.com/opensearch-project/opensearch-api-specification as well and add tests there for this and other use-cases so we don't lose this functionality as we move towards more of this code being generated.

CHANGELOG.md
```diff
@@ -9,6 +9,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Removed
 ### Fixed
 - Fixed Search helper to ensure proper retention of the _collapse attribute in chained operations. ([#771](https://github.com/opensearch-project/opensearch-py/pull/771))
+- Fixed the use of `minimum_should_match` with `Bool` to allow the use of string-based value (percent string, combination)
```
dblock, line 12: Add PR number here.

opensearchpy/helpers/query.py
```diff
@@ -219,10 +220,17 @@ def __and__(self, other: "Bool") -> Any:
                 del q._params["minimum_should_match"]
 
             for qx in (self, other):
-                # TODO: percentages will fail here
                 min_should_match = qx._min_should_match
+
+                # attempt to convert a string integer representation to int
+                with contextlib.suppress(ValueError):
```
dblock, line 226: I am worried that `int()` may do unexpected things like rounding a float for example. 

Maybe we can just be explicit, check that the value ends with % and convert the rest?

dblock (commented):
Looking pretty good, thanks for the tests. See below for comments.

Please do fix DCO, `git commit --amend -s` and force push will do it.

opensearchpy/helpers/query.py
dblock: I think you misunderstood me, sorry. 

I mean that if we rely on an exception to see if the value can be converted, then we might introduce undesired behavior. 

```
$ python -c "print(int("3.14"))"
3
```

You can see that if I pass an incorrect value it's happily coerced into an int, losing the actual value! That doesn't seem good :)

dblock (approved):

dblock:
Thanks for fixing this!

PR #781, opened 2024-07-20
Remove redundant six dependency

hugovk:
### Description
_Describe what this change achieves._

This package only supports Python 3, however it still uses the [six](https://six.readthedocs.io/) compatibility library to support Python 2 and 3.

This PR removes the dependency, and replaces the six calls with direct Python 3 code.

This makes the install a bit quicker, smaller, and the code a bit faster by using the standard library instead of a package.

### Issues Resolved

None known.

---

I did this using the https://github.com/asottile/pyupgrade tool that upgrades Python syntax.

We could run it on all the files, and upgrade to Python 3.8+ syntax (`--py38-plus`), but I wanted to keep the diff small, so I only ran it on files using six, and did the lowest syntax upgrade (`--py3-plus`):

```sh
rg six -l | xargs pyupgrade --py3-plus
```

(I do recommend running it on the whole codebase and for 3.8+, perhaps in another PR. For example `pyupgrade **/*.py --py38-plus` upgrades 84 further files.)

---

This PR also removes creation of universal wheels, that's only needed when supporting Python 2, and corrects the Black target version.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Removed deprecated `numpy.float_` and update NumPy/Pandas imports ([#762](https://github.com/opensearch-project/opensearch-py/pull/762))
 ### Deprecated
 ### Removed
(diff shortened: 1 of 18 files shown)


dblock (approved):
Looks good, thank you! Would definitely appreciate future upgrades.

PR #789, opened 2024-07-30
Fix spell typos and missing whitespace

m3et:
### Description
Fix missing whitespaces in IndicesClient docstrings

### Issues Resolved
Did not open an issue

### DIFF
--- a/opensearchpy/client/indices.py
+++ b/opensearchpy/client/indices.py
@@ -335,7 +335,7 @@ class IndicesClient(NamespacedClient):
             index aliases used to limit the request. Wildcard expressions (*) are
             supported.
         :arg allow_no_indices: If false, the request returns an error if
-            any wildcard expression, index alias, or _all value targets onlymissing
+            any wildcard expression, index alias, or _all value targets only missing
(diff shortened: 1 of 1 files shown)


dblock:
Is this code generated from spec (it should be)? I am afraid the bug is in the generator or the spec.

https://github.com/opensearch-project/opensearch-api-specification

dblock (approved):

dblock:
@m3et First, I merged this PR, thank you.

I found other instances of the same in another client, https://github.com/search?q=org%3Aopensearch-project%20onlymissing&type=code, if you are interested in fixing those.

> Can you elaborate on that? Does it mean the the project doc is generated from this project and I should look for the bug there?

Not yet, but it will be.  We will need to merge descriptions between what's here and what's in the spec.

dblock:
This was a bug in the generator, the next run undid all these changes. 

I fixed it in https://github.com/opensearch-project/opensearch-py/pull/793/files#diff-999765094c323a96b13cc24c24aa13948be15e6beacda7747e83c300420f8018R626.

PR #804, opened 2024-08-22
Add override for code generator to change `indices.put_alias` argument order

### LINKED ISSUE #803: [BUG] v2.7.0 brings undocumented breaking change to put_index()
### What is the bug?
With versions prior to version 2.7.0 it was possible to call `client.indices.put_alias("index_name", "alias_name")` and have it create the alias.

With version 2.7.0 you instead get a cryptic error message along the lines of:

> opensearchpy.exceptions.RequestError: RequestError(400, 'json_parse_exception', "Unrecognized token 'alias_name': was expecting (JSON String, Number, Array, Object or token 'null', 'true' or 'false')\n at [Source: REDACTED (`StreamReadFeature.INCLUDE_SOURCE_IN_LOCATION` disabled); line: 1, column: 13]")

1. Create an index
2. Attempt to create an alias for it with `client.indices.put_alias("index_name", "alias_name")`

The alias should be created for the index like in previous opensearch-py versions.

N/A

N/A

This is caused by an inadvertently breaking change in https://github.com/opensearch-project/opensearch-py/pull/777 which changed the order of the arguments to the `put_alias()` method, meaning anybody who called the method without naming the arguments now supply their values to different arguments.

Tenzer:
### Description
This solves an issue introduced in v2.7.0 where the argument order of the `indices.put_alias()` method were altered, resulting in a breaking change for any user that called the function without naming the arguments provided.

### Issues Resolved
Closes #803.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -24,6 +24,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 - Fixed the use of `minimum_should_match` with `Bool` to allow the use of string-based value (percent string, combination). ([#780](https://github.com/opensearch-project/opensearch-py/pull/780))
 - Fixed incorrect `retry_on_conflict` type ([#795](https://github.com/opensearch-project/opensearch-py/pull/795))
 ### Updated APIs
+- Updated opensearch-py APIs to reflect [opensearch-api-specification@b28c1cf](https://github.com/opensearch-project/opensearch-api-specification/commit/b28c1cfaf4a76d786c789929eaa7fabffd09cb99)
 - Updated opensearch-py APIs to reflect [opensearch-api-specification@9d3bc34](https://github.com/opensearch-project/opensearch-api-specification/commit/9d3bc340ccd7d049e7d6e14a4aff2293780cb446)
 ### Security
 ### Dependencies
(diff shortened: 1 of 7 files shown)


dblock (approved):
Works for me! Thanks for fixing this.

dblock:
@Tenzer did you happen to go through a diff of other APIs that might have this problem? Maybe there's a way to do this systematically 🤔  I want to double check before cutting a newer release.

DarshitChanpura:
@dblock this is interesting, post is exposed but doesn't really consume any body for the request.
https://github.com/opensearch-project/security/blob/main/src/main/java/org/opensearch/security/rest/DashboardsInfoAction.java#L94
We only seem to be using get request:
https://github.com/opensearch-project/security-dashboards-plugin/blob/main/server/routes/index.ts#L584
https://github.com/opensearch-project/security-dashboards-plugin/blob/main/public/utils/dashboards-info-utils.tsx#L21-L34

Also, could you point the exact line change for post_dashboards_info that might be causing this. https://github.com/opensearch-project/opensearch-py/compare/v2.6.0...v2.7.0#diff-d373e8c1b532ced908538444607c89e5a80cfe7a24ef5c81a3708e0fb2cf48fdL2173 doesn't seem to match any text when searching for "dashboards_info". Maybe I missed something?

dblock:
@DarshitChanpura Thanks. I think we're all good, since this method doesn't take a `body` it was a bug fix.

I am sure where this is evaluated in the client code generator, but you can dig through https://github.com/opensearch-project/opensearch-py/blob/main/utils/generate_api.py.

Issue #821, opened 2024-09-16
[BUG] The proxy setup does not work with Urllib3HttpConnection


int0x03:
### What is the bug?
the proxy setup in this wiki does not work https://github.com/opensearch-project/opensearch-py/blob/main/guides/proxy.md

### How can one reproduce the bug?
use the code in wiki

### What is the expected behavior?
follow the wiki, the proxy should work

### What is your host/environment?
MAC

### Do you have any screenshots?
no

### Do you have any additional context?
no

florian-lang-a1:
If you are using the code for the Sync Client there is actually a line missing there.
It should be
```python
from opensearchpy import OpenSearch, RequestsHttpConnection

OpenSearch(
    hosts=["https://..."],
    use_ssl=True,
    verify_certs=True,
    connection_class=RequestsHttpConnection,
    trust_env=True,
)
```

dblock:
@int0x03 @florian-lang-a1 care to help fix the doc please?

int0x03:
thanks @florian-lang-a1  @dblock . 
with @florian-lang-a1 suggestion, it means we must setup the proxy in environment variables right? if yes, it means the proxy is used by all the client connection code. that's the not correct way. 

it is better to have a proxies parameter, like the python requests:

```
import requests

proxies = {
    'http': 'http://proxy.example.com:8080',
    'https': 'http://proxy.example.com:8080',
}

response = requests.get('http://example.com', proxies=proxies)
print(response.text)
```

dblock:
> it is better to have a proxies parameter, like the python requests:

Yes, you're correct.

int0x03:
I will check how to add this parameter and send a PR in the weekend.

dblock:
Looks like `RequestsHttpConnection` works for the proxy scenario but not `Urllib3HttpConnection`. This is the bug that needs to be fixed.

PR #837, opened 2024-10-23
fix proxy.md

jiangying000:
related: https://github.com/opensearch-project/opensearch-py/issues/821

### DIFF
--- a/guides/proxy.md
+++ b/guides/proxy.md
@@ -10,9 +10,10 @@
 from opensearchpy import OpenSearch, RequestsHttpConnection
 
 OpenSearch(
-    hosts=["htps://..."],
+    hosts=["https://..."],
     use_ssl=True,
     verify_certs=True,
+    connection_class=RequestsHttpConnection,
     trust_env=True,
 )
 ```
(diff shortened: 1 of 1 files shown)


dblock:
Why doesn't it work with the default connection class?

dblock:
> RequestsHttpConnection

There are multiple connection classes, see https://github.com/opensearch-project/opensearch-py/blob/9dd7bf7cb3e80b98fcc256edb9c49cf89202f522/guides/connection_classes.md.

The default is `Urllib3HttpConnection` for synchronous scenarios. The change you are proposing switching a non-working proxy implementation in `Urllib3HttpConnection` to a working proxy implementation in `RequestsHttpConnection`, so I don't think this is the right fix, aside of the documentation someone needs to actually fix the problem for proxy support in `Urllib3HttpConnection`.

jiangying000:
Agree, this is not a fix to #821 , just a temporary fix to the doc

At least Users can get what they want by following the doc with this change

dblock (approved):

dblock:
I agree, I'll merge this.

Issue #833, opened 2024-10-17
[BUG] Bug with `AWS` if `id` has special characters


Danipulok:
### What is the bug?
AWS + special characters in 'id' result in:
```
opensearchpy.exceptions.AuthorizationException: AuthorizationException(403, '{"message":"The request signature we calculated does not match the signature you provided. Check your AWS Secret Access Key and signing method. Consult the service documentation for details.\\n\\nThe Canonical String for this request should have been\\n\'GET\\n/index-name/_doc/cool_bot%21-f9aa36ea-b8a5-4890-957e-29f000beec86%40localhost\\n\\nhost:search.us-west-2.es.amazonaws.com\\nx-amz-date:20241017T115545Z\\n\\nhost;x-amz-date\\ne3b0c4429\'\\n\\nThe String-to-Sign should have been\\n\'AWS4-HMAC-SHA256\\n20241017T115545Z\\n20241017/us-west-2/es/aws4_request\\n262223abc008b0d31e74a714950\'\\n"}')
```

### How can one reproduce the bug?
```python
import boto3
from opensearchpy import AWSV4SignerAsyncAuth, AsyncHttpConnection, AsyncOpenSearch

OPENSEARCH_URL = "https://search.us-west-2.es.amazonaws.com"
OPENSEARCH_AWS_ACCESS_KEY_ID = "..."
OPENSEARCH_AWS_SECRET_ACCESS_KEY = "..."
OPENSEARCH_AWS_REGION = "us-west-2"

kwargs = {}
credentials = boto3.Session(
    aws_access_key_id=OPENSEARCH_AWS_ACCESS_KEY_ID,
    aws_secret_access_key=OPENSEARCH_AWS_SECRET_ACCESS_KEY,
).get_credentials()
auth = AWSV4SignerAsyncAuth(credentials, OPENSEARCH_AWS_REGION)
kwargs["http_auth"] = auth
kwargs["connection_class"] = AsyncHttpConnection


async def main() -> None:
    opensearch_client = AsyncOpenSearch(
        OPENSEARCH_URL,
        **kwargs,
    )

    REAL_ID = "cool_bot@localhost"
    await opensearch_client.get(index="index_name", id=REAL_ID)

    await opensearch_client.close()


if __name__ == "__main__":
    asyncio.run(main())
```

### What is the expected behavior?
When using localhost openseach run via docker everything's okay. So I guess there's a problem with `AWSV4SignerAsyncAuth` somewhere

[...]

PR #839, opened 2024-10-28
Fix bug where URL is being signed after being quoted

fabiopedrosa:
### Description

URL is being signed after being quoted, but should be before.

### Issues Resolved
Closed https://github.com/opensearch-project/opensearch-py/issues/833

### DIFF
--- a/opensearchpy/helpers/asyncsigner.py
+++ b/opensearchpy/helpers/asyncsigner.py
@@ -8,6 +8,7 @@
 # GitHub history for details.
 
 from typing import Any, Dict, Optional, Union
+from ..compat import unquote
 
 
 class AWSV4SignerAsyncAuth:
@@ -56,7 +57,7 @@ class AWSV4SignerAsyncAuth:
         # create an AWS request object and sign it using SigV4Auth
         aws_request = AWSRequest(
             method=method,
-            url=url,
+            url=unquote(url),
             data=body,
         )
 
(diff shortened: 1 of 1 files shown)


dblock (changes requested):
I don't think this is the right fix. The URL could have been already quoted before, in which case it will start failing again. Look for where it gets quoted in the first place, and preserve the original URL to be used here for signing. 

Start by writing a failing test, too, and let's make sure it exists for all transport classes.

fabiopedrosa:
@dblock I agree, I just won't have time to do that work right now.

dblock:
> @dblock I agree, I just won't have time to do that work right now. Is it something AWS can contribute to?

We can see. Do you have a ticket with customer support by any chance?

besil:
hello, I just opened a case to AWS support before discovering this thread. If you need it, I can provide the AWS Support Ticket ID

mingshl:
it seems this ticket is out of dated. can you please close it ?@fabiopedrosa

dblock:
Fixed in https://github.com/opensearch-project/opensearch-py/pull/848.

dblock:
@fabiopedrosa If you're relying on the fix you proposed here I suggest reverting to HEAD (and the next release of opensearch-py). This fix doesn't work in half the cases with indices with special characters for example.

fabiopedrosa:
@dblock just tried using HEAD and works great. thank you a lot for the fixes, and thank you @nathaliellenaa

PR #848, opened 2024-11-19
Fix AuthorizationException with AWSV4SignerAsyncAuth when the doc ID has special characters.

nathaliellenaa:
### Description
1. Fix for https://github.com/opensearch-project/opensearch-py/issues/833.
2. Added tests for `_make_path`.
3. Added integration tests from https://github.com/opensearch-project/opensearch-py/pull/856.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Fixed
 - Fix `Transport.perform_request`'s arguments `timeout` and `ignore` variable usage ([810](https://github.com/opensearch-project/opensearch-py/pull/810))
 - Fix `Index.save` not passing through aliases to the underlying index ([823](https://github.com/opensearch-project/opensearch-py/pull/823))
+- Fix `AuthorizationException` with AWS OpenSearch when the doc ID contains `:` ([848](https://github.com/opensearch-project/opensearch-py/pull/848))
 ### Security
 
 ### Dependencies
--- a/test_opensearchpy/test_async/test_signer.py
+++ b/test_opensearchpy/test_async/test_signer.py
@@ -8,6 +8,7 @@
 # GitHub history for details.
 
 import uuid
+from typing import Any, Collection, Dict, Mapping, Optional, Tuple, Union
 from unittest.mock import Mock
 
 import pytest
@@ -103,3 +104,75 @@ class TestAsyncSignerWithFrozenCredentials(TestAsyncSigner):
         assert "X-Amz-Date" in headers
         assert "X-Amz-Security-Token" in headers
         assert len(mock_session.get_frozen_credentials.mock_calls) == 1
+
+
+class TestAsyncSignerWithSpecialCharacters:
+    def mock_session(self) -> Mock:
+        access_key = uuid.uuid4().hex
+        secret_key = uuid.uuid4().hex
+        token = uuid.uuid4().hex
+        dummy_session = Mock()
+        dummy_session.access_key = access_key
+        dummy_session.secret_key = secret_key
+        dummy_session.token = token
+
+        del dummy_session.get_frozen_credentials
+
+        return dummy_session
+
+    async def test_aws_signer_async_consitent_url(self) -> None:
+        region = "us-west-2"
+
+        from opensearchpy import AsyncOpenSearch
+        from opensearchpy.connection.http_async import AsyncHttpConnection
+        from opensearchpy.helpers.asyncsigner import AWSV4SignerAsyncAuth
+
+        # Store URLs for comparison
+        signed_url = None
+        sent_url = None
+
+        doc_id = "doc_id:with!special*chars%3A"
+        quoted_doc_id = "doc_id%3Awith%21special*chars%253A"
+        url = f"https://search-domain.region.es.amazonaws.com:9200/index/_doc/{quoted_doc_id}"
+
+        # Create a mock signer class to capture the signed URL
+        class MockSigner(AWSV4SignerAsyncAuth):
+            def _sign_request(
+                self,
+                method: str,
+                url: str,
+                query_string: Optional[str] = None,
+                body: Optional[Union[str, bytes]] = None,
+            ) -> Dict[str, str]:
+                nonlocal signed_url
+                signed_url = url
+                return {}
+
+        # Create a mock connection class to capture the sent URL
+        class MockConnection(AsyncHttpConnection):
+            async def perform_request(
+                self: "MockConnection",
+                method: str,
+                url: str,
+                params: Optional[Mapping[str, Any]] = None,
+                body: Optional[Any] = None,
+                timeout: Optional[Union[int, float]] = None,
+                ignore: Collection[int] = (),
+                headers: Optional[Mapping[str, str]] = None,
+            ) -> Tuple[int, Mapping[str, str], str]:
+                nonlocal sent_url
+                sent_url = f"{self.host}{url}"
+                return 200, {}, "{}"
+
+        auth = MockSigner(self.mock_session(), region)
+        auth("GET", url)
+
+        client = AsyncOpenSearch(
+            hosts=[{"host": "search-domain.region.es.amazonaws.com"}],
+            http_auth=auth,
+            use_ssl=True,
+            verify_certs=True,
+            connection_class=MockConnection,
+        )
+        await client.index("index", {"test": "data"}, id=doc_id)
+        assert signed_url == sent_url, "URLs don't match"
--- a/DEVELOPER_GUIDE.md
+++ b/DEVELOPER_GUIDE.md
@@ -101,7 +101,11 @@ test_opensearchpy/test_connection.py::TestRequestsConnection::test_no_http_compr
 test_opensearchpy/test_async/test_connection.py::TestAIOHttpConnection::test_no_http_compression PASSED [100%]
 ```
 
-Note that integration tests require docker to be installed and running, and downloads quite a bit of data from over the internet and hence take few minutes to complete.
+```
+./.ci/run-tests false 2.16.0 test_indices_lifecycle
+```
+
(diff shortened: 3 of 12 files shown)


test_opensearchpy/test_async/test_signer.py
```diff
@@ -103,3 +103,48 @@ async def test_aws_signer_async_frozen_credentials_as_http_auth(self) -> None:
...
+        # Test with unencoded ID
+        url_raw = f"https://search-domain.region.es.amazonaws.com/index/_doc/{doc_id}"
+        headers_raw = auth("GET", url_raw)
+        
+        # Test with properly encoded ID
+        encoded_id = quote(doc_id, safe='')
+        url_encoded = f"https://search-domain.region.es.amazonaws.com/index/_doc/{encoded_id}"
+        headers_encoded = auth("GET", url_encoded)
+        
+        # Signatures should be the same since URL encoding happens before signing
```
dblock, line 142: Isn't it that currently the URL we send is not the URL we use for signing?

If so, we don't want the signing code to be decoding previously encoded URLs or knowing anything about URLs. The correct behavior would be that the URL passed into the client for an `index/_doc/{doc_id}` path is also the url that is used for signing. If we do any URL encoding, it should happen before signing, and then the HTTP request to the server uses the encoded version, too (so that the server can check it).

nathaliellenaa: Hmm that makes sense, I'll fix the test.

dblock (changes requested):
Thanks. 

The responsibility of the signer is to sign data blindly, ie. it's not supposed to know anything about encoding/decoding. This implementation will fail if I actually encode the URL parameter on my own, then pass it into the client.

Look for a test that ensures that the URL being sent is the same URL that is being signed, and a fix where the URL is encoded before both sending and signing.

dblock (commented):
Other than the fix, see if you can simplify this test further. It's likely that you don't need any fields in the mock session because you're now stubbing signing altogether and thus you can get rid of `super().sign_request ...` in the mock signer. This also means you don't need the connection because you won't be making actual requests. Finally, don't rely on general exceptions in the test. Your mock signer could raise an expected error, but it would be best if you replace the connection with a mock that just returns without doing anything.

dblock (changes requested):
The test looks ok, but this is not the right fix.

1.  If the DOC ID is `foo%3Abar` this code will decode it to `foo:bar` and fail the same way. We encode the URL somewhere, find where, save it for the signer before it's being encoded, then pass the original url into the signer.
2. The same problem (likely) exists in the synchronous path, add tests and a fix in both.

CHANGELOG.md
```diff
@@ -12,6 +12,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Fixed
 - Fix `Transport.perform_request`'s arguments `timeout` and `ignore` variable usage ([810](https://github.com/opensearch-project/opensearch-py/pull/810))
 - Fix `Index.save` not passing through aliases to the underlying index ([823](https://github.com/opensearch-project/opensearch-py/pull/823))
+- Fix bug where the URL being sent and being signed is different ([848](https://github.com/opensearch-project/opensearch-py/pull/848))
```
dblock, line 15:
```suggestion
- Fix `AuthorizationException` with AWS OpenSearch when the doc ID contains `:` ([848](https://github.com/opensearch-project/opensearch-py/pull/848))
```

dblock:
Much better! Get CI to pass.

I tested this using https://github.com/dblock/opensearch-python-client-demo.

```
osx ~/source/dblock/opensearch-python-client-demo/async (main)$ git diff example.py
diff --git a/async/example.py b/async/example.py
index 2360ebf..1754609 100644
--- a/async/example.py
+++ b/async/example.py
@@ -55,19 +55,23 @@ async def main():
 
   try:
     # index data
+    id = 'id@local'
     document = {'director': 'Bennett Miller', 'title': 'Moneyball', 'year': 2011}
-    await client.index(index=index, body=document, id='1')
+    await client.index(index=index, body=document, id=id)
 
     # wait for the document to index
     sleep(1)
 
+    doc = await client.get(index=index, id=id)
+    print(doc)
+
     # search for the document
     results = await client.search(body={'query': {'match': {'director': 'miller'}}})
     for hit in results['hits']['hits']:
       print(hit['_source'])
 
     # delete the document
-    await client.delete(index=index, id='1')
+    await client.delete(index=index, id=id)
   finally:
     # delete the index
     await client.indices.delete(index=index)
```
```
osx ~/source/dblock/opensearch-python-client-demo/async (main)$ git diff Pipfile
diff --git a/async/Pipfile b/async/Pipfile
index f2923c4..35cb795 100644
--- a/async/Pipfile
+++ b/async/Pipfile
@@ -4,10 +4,10 @@ verify_ssl = true
 name = "pypi"
 
 [packages]
-opensearch-py = "2.4.1"
+opensearch-py = {git = "git+https://github.com/nathaliellenaa/opensearch-py.git", ref = "python-client-bug"}
 boto3 = "*"
 
 [dev-packages]
 
 [requires]
-python_version = "3"
\ No newline at end of file
+python_version = "3"
```

Failed with 2.4.1 and worked great with your changes.

opensearchpy/_async/transport.py
```diff
@@ -397,7 +399,8 @@ async def perform_request(
                     ignore=ignore,
                     timeout=timeout,
                 )
-
+                url = quote(url)
```
dblock, line 402: This is done in a loop `for attempt ...`, might fail the second time quoting an already quoted URL.

nathaliellenaa:
> Failed with 2.4.1 and worked great with your changes.
Do you know why it fails on some specific versions?

dblock:
> > Failed with 2.4.1 and worked great with your changes.
> 
> Do you know why it fails on some specific versions?

Looks like legit test failures that have to do with the changes to URL encoding.

```
=========================== short test summary info ============================
FAILED test_opensearchpy/test_client/test_utils.py::TestMakePath::test_handles_unicode - AssertionError: '/some-index/type/%E4%B8%AD%E6%96%87' != '/some-index/type/中文'
- /some-index/type/%E4%B8%AD%E6%96%87
+ /some-index/type/中文
```

nathaliellenaa:
> Looks like legit test failures that have to do with the changes to when URLs get encoded.
> 
> ```
> =========================== short test summary info ============================
> FAILED test_opensearchpy/test_client/test_utils.py::TestMakePath::test_handles_unicode - AssertionError: '/some-index/type/%E4%B8%AD%E6%96%87' != '/some-index/type/中文'
> - /some-index/type/%E4%B8%AD%E6%96%87
> + /some-index/type/中文
> ```

Yes, do I need to modify this test since the behavior of the `_make_path` function has been modified?

dblock:
> Yes, do I need to modify this test since the behavior of the `_make_path` function has been modified?

Correct.

dblock:
Tested with samples from #857.

```
osx ~/source/opensearch-project/opensearch-py/dblock-opensearch-py/samples (nathaliellenaa-python-client-bug)$ poetry run python aws/search_urllib3_async.py 
INFO:Found credentials in environment variables.
INFO:GET https://search-dblock-test-opensearch-27-xwd6xrwtykennk3q7uu7t7dtpi.us-west-2.es.amazonaws.com:443/ [status:200 request:0.314s]
opensearch: 2.7.0
INFO:PUT https://search-dblock-test-opensearch-27-xwd6xrwtykennk3q7uu7t7dtpi.us-west-2.es.amazonaws.com:443/movies [status:200 request:0.576s]
INFO:PUT https://search-dblock-test-opensearch-27-xwd6xrwtykennk3q7uu7t7dtpi.us-west-2.es.amazonaws.com:443/movies/_doc/foo%40bar [status:201 request:0.188s]
INFO:POST https://search-dblock-test-opensearch-27-xwd6xrwtykennk3q7uu7t7dtpi.us-west-2.es.amazonaws.com:443/_search [status:200 request:0.102s]
{'director': 'Bennett Miller', 'title': 'Moneyball', 'year': 2011}
INFO:DELETE https://search-dblock-test-opensearch-27-xwd6xrwtykennk3q7uu7t7dtpi.us-west-2.es.amazonaws.com:443/movies/_doc/foo%40bar [status:200 request:0.116s]
INFO:DELETE https://search-dblock-test-opensearch-27-xwd6xrwtykennk3q7uu7t7dtpi.us-west-2.es.amazonaws.com:443/movies [status:200 request:0.153s]
osx ~/source/opensearch-project/opensearch-py/dblock-opensearch-py/samples (nathaliellenaa-python-client-bug)$ 
```

```
poetry run python aws/search_urllib3_sync.py 
INFO:Found credentials in environment variables.
INFO:GET https://search-dblock-test-opensearch-27-xwd6xrwtykennk3q7uu7t7dtpi.us-west-2.es.amazonaws.com:443/ [status:200 request:0.321s]
opensearch: 2.7.0
INFO:PUT https://search-dblock-test-opensearch-27-xwd6xrwtykennk3q7uu7t7dtpi.us-west-2.es.amazonaws.com:443/movies [status:200 request:0.445s]
INFO:PUT https://search-dblock-test-opensearch-27-xwd6xrwtykennk3q7uu7t7dtpi.us-west-2.es.amazonaws.com:443/movies/_doc/foo%40bar [status:201 request:0.179s]
INFO:POST https://search-dblock-test-opensearch-27-xwd6xrwtykennk3q7uu7t7dtpi.us-west-2.es.amazonaws.com:443/_search [status:200 request:0.100s]
{'director': 'Bennett Miller', 'title': 'Moneyball', 'year': 2011}
INFO:DELETE https://search-dblock-test-opensearch-27-xwd6xrwtykennk3q7uu7t7dtpi.us-west-2.es.amazonaws.com:443/movies/_doc/foo%40bar [status:200 request:0.099s]
INFO:DELETE https://search-dblock-test-opensearch-27-xwd6xrwtykennk3q7uu7t7dtpi.us-west-2.es.amazonaws.com:443/movies [status:200 request:0.149s]
```

dblock (approved):

fabiopedrosa (approved):

PR #851, opened 2024-11-25
Fix `/cluster/stats` in a backwards compatible way 

dblock:
### Description

The `/_cluster/stats` endpoint exists now in all its glorious variations up to `/_cluster/stats/indices/analysis/nodes/test-node` which includes a metric (new), an index metric (new), and a node id (old). The generator added the metric and index metric in order, added a patch that swaps them to the end of the list to preserve backwards compatibility.

Adds `insights`, `list` and `wlm` APIs amongst others from spec.


saimedhi (approved):
LGTM.

Issue #337, opened 2023-03-21
[BUG] search method doesn't support search_after parameter


r-mendes:
### What is the bug?
The `search` method doesn't support `search_after` parameter.

### How can one reproduce the bug?
1. Instantiate a OpenSearch object;
2. Run a search using the `sort` parameter;
3. Get the result and then look for the last `hit` entry;
4. From the last `hit` entry, fetch the `sort` field value;
5. Run a `search` passing the value fetched in step 4 in the `searh_after` parameter;
6. It will be shown a messaging informing that this parameter is not supported;

### What is the expected behavior?
The `search_after` parameter supposed to be supported in the most recent version of the library.

### What is your host/environment?
Linux, Ubuntu 22.04

### Do you have any screenshots?
- ElasticSearch library source code: https://github.dev/elastic/elasticsearch-py/blob/20133a338e01b0e6073ccd90bf154a5364df0ca4/elasticsearch/_sync/client/__init__.py#L3476
- OpenSearch library source code: https://github.dev/opensearch-project/opensearch-py/blob/a495a20417ea375abd9421dd54c0d3e04c586354/opensearchpy/client/__init__.py#L1520

### Do you have any additional context?
The lack of this parameter block the usage of queries' pagination using the `search_after` and `Point-in-time` approach.

margulanz:
I will try to work on this issue

ReinGrad:
The OpenSearch library currently does not support the` search_after` parameter, which is used for pagination in search queries. This means that users cannot easily paginate large result sets using the `search_after` parameter in OpenSearch.

It would be possible that the OpenSearch library should support the `search_after` parameter, since this is a common feature in other search libraries such as ElasticSearch. It is possible that this feature may be added in a future version of the library.

One of the options may be switching to another search library that supports the search_after parameter, or implementing a different pagination strategy using the from and size parameters. It may also be possible to contribute to the OpenSearch library to add support for the `search_after `parameter.

margulanz:
As I understand, this issue may be considered as closed? @ReinGrad?

wbeckler:
If anyone wants to add this as a feature, I think it still makes sense. Or if anyone wants to update the user guide examples for now.

dblock:
@r-mendes or @morrissimo Any interest in picking this up?

dblock:
[...]

PR #859, opened 2024-11-29
Added a support for search (`search_after` parameter)

nathaliellenaa:
### Description
Added sync and async sample that uses `search_after` parameter for https://github.com/opensearch-project/opensearch-py/issues/337. 


dblock:
One more thing. Maybe we can write a sample that inserts many records and paginates through N pages instead until there are no more results? That seems more useful than a synthetic example with just 3 values.

dblock (approved):

PR #863, opened 2024-11-30
made custom headers be available to async aws signer

### LINKED ISSUE #184: [BUG] Allow setting alternate host header (with port) in AWSV4SignerAuth

brunomurino:
### Description
When creating an Async client, you can now pass a custom 'host' header that will be used for signing the AWS request, if using AWS authentication. This allows accessing OpenSearch via SSH/SSM tunnel, since when doing so the local port of OS would be, most of the time, 443 which is protected.

### Issues Resolved
Closes #184 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 ### Added
+- Added option to pass custom headers to 'AWSV4SignerAsyncAuth' ([863](https://github.com/opensearch-project/opensearch-py/pull/863))
 ### Updated APIs
 ### Changed
 ### Deprecated
--- a/guides/auth.md
+++ b/guides/auth.md
@@ -2,6 +2,7 @@
   - [IAM Authentication](#iam-authentication)
   - [IAM Authentication with a Synchronous Client](#iam-authentication-with-a-synchronous-client)
   - [IAM Authentication with an Async Client](#iam-authentication-with-an-async-client)
+  - [IAM Authentication via Tunnel](#iam-authentication-via-tunnel)
   - [Kerberos](#kerberos)
 
 # Authentication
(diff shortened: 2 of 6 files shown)


brunomurino:
@dblock Indeed the Async transport wasn't leveraging the custom headers passed, so I added that following how it was done in the sync version. Also added a couple of tests.

I couldn't find where to add public documentation. I'm happy to add some bits if you point me to the right place.

Also, the coverage test is failing but I'm not sure it's due to my changes?

brunomurino:
Added some docs in the existing `auth.md` guide

dblock:
> Also, the coverage test is failing but I'm not sure it's due to my changes?

Likely not. Would be nice if someone got to the bottom of it.

dblock (approved):
Thanks for adding this!

Two nice to have things for a future PR if you want to improve it.

1. `_fetch_url` is common for 2 signers, refactor either into a base class or a utility
2. The test "knows" that there's an internal `_fetch_url` which is an antipattern, it should be doing the signing and ensuring the signer is called with a different URL - I think we have other tests that use a mock signer already if you want to copy from

brunomurino:
Great! Thanks for approving! I agree with the suggested improvements.I did try other test approaches but ended up with using the internal function. I’ll look into the mock signer!On 1 Dec 2024, at 13:32, Daniel (dB.) Doubrovkine ***@***.***> wrote:﻿
@dblock approved this pull request.

Thanks for adding this!
Two nice to have things for a future PR if you want to improve it.

_fetch_url is common for 2 signers, refactor either into a base class or a utility
The test "knows" that there's an internal _fetch_url which is an antipattern, it should be doing the signing and ensuring the signer is called with a different URL - I think we have other tests that use a mock signer already if you want to copy from

brunomurino:
@dblock Regarding the test, would you be opposed to using something like:
```python
with patch("botocore.awsrequest.AWSRequest", side_effect=AWSRequest) as mock_AWSRequest:
```
to see if the right argument was passed? would that still be an antipattern?

dblock:
I think that makes things a lot simpler!

Issue #267, opened 2023-01-11
[BUG] CI tests are not working locally


saimedhi:
### What is the bug?
CI tests are not working locally.

### What is the expected behavior?
Fix the CI tests such that it works locally.

saimedhi:
Working on it.

saimedhi:
`nox > Running session test-3.9
nox > Re-using existing virtual environment at .nox/test-3-9.
nox > python -m pip install .
nox > python -m pip install -r dev-requirements.txt
nox > python setup.py test
running test
WARNING: Testing via this command is deprecated and will be removed in a future version. Users looking for a generic test entry point independent of test runner are encouraged to use tox.
/Users/saimedhi/code3/opensearch-py/.nox/test-3-9/lib/python3.9/site-packages/setuptools/installer.py:27: SetuptoolsDeprecationWarning: setuptools.installer is deprecated. Requirements should be satisfied by a PEP 517 installer.
  warnings.warn(
running egg_info
writing opensearch_py.egg-info/PKG-INFO
writing dependency_links to opensearch_py.egg-info/dependency_links.txt
writing requirements to opensearch_py.egg-info/requires.txt
writing top-level names to opensearch_py.egg-info/top_level.txt
reading manifest file 'opensearch_py.egg-info/SOURCES.txt'
reading manifest template 'MANIFEST.in'
warning: no files found matching 'Changelog.rst'
warning: no files found matching 'LICENSE'
no previously-included directories found matching 'test_opensearch'
warning: no previously-included files matching '__pycache__' found under directory '*'
warning: no previously-included files matching '*.py[co]' found under directory '*'
adding license file 'LICENSE.txt'
adding license file 'NOTICE.txt'
adding license file 'AUTHORS'
writing manifest file 'opensearch_py.egg-info/SOURCES.txt'
running build_ext
GET https://localhost:9200/ [status:N/A request:0.029s]
Traceback (most recent call last):
  File "/Users/saimedhi/code3/opensearch-py/.nox/test-3-9/lib/python3.9/site-packages/urllib3/connectionpool.py", line 703, in urlopen
    httplib_response = self._make_request(
  File "/Users/saimedhi/code3/opensearch-py/.nox/test-3-9/lib/python3.9/site-packages/urllib3/connectionpool.py", line 386, in _make_request
    self._validate_conn(conn)
[...]

PR #879, opened 2024-12-27
Fixed test client initialization in fetching to ensure tests run locally

nokados:
### Description
It fixes the same problem as in #267 in the same way, but this time it was caused by `fetch_opensearch_repo`

### DIFF
--- a/test_opensearchpy/run_tests.py
+++ b/test_opensearchpy/run_tests.py
@@ -68,7 +68,10 @@ def fetch_opensearch_repo() -> None:
 
     # find out the sha of the running client
     try:
-        client = get_client()
+        client = get_client(
+            verify_certs=False,
+            http_auth=("admin", environ.get("OPENSEARCH_PASSWORD", "admin"))
+        )
         sha = client.info()["version"]["build_hash"]
     except (SkipTest, KeyError):
         print("No running opensearch >1.X server...")


dblock (approved):

dblock:
There are two failing test suites. Probably unrelated, but care to take a look?

dblock (changes requested):
There are two failing test suites with this change, care to take a look?

dblock:
Hmmm, this seems like a workaround around a workaround. I see a lot of code like [this](https://github.com/opensearch-project/opensearch-py/blob/4cdf81e74142ddd2f501518b54ecb9591fbd873a/test_opensearchpy/test_async/test_server/__init__.py#L42)

Feels like a fix should be somewhere deeper. 

What are the state(s) of your environment variables `OPENSEARCH_INITIAL_ADMIN_PASSWORD` and `OPENSEARCH_PASSWORD` that causes a local run to fail?

nokados:
To reproduce the issue, just follow the [DEVELOPER_GUIDE](https://github.com/opensearch-project/opensearch-py/blob/main/DEVELOPER_GUIDE.md). 

So, in the local environment, OPENSEARCH_PASSWORD may be set. OPENSEARCH_INITIAL_ADMIN_PASSWORD is for [integration tests](https://github.com/opensearch-project/opensearch-py/blob/4cdf81e74142ddd2f501518b54ecb9591fbd873a/.github/workflows/integration.yml#L25). 

I’m currently unable to dedicate time to explore a deeper solution. At the moment, the code isn’t working as intended. If the extra two lines in the tests feel unnecessary to you, you’re welcome to refactor it, but unfortunately, I won’t be able to take that on right now.

dblock:
Thanks for your help so far @nokados.

PR #971, opened 2025-11-04
Preparing for release, 3.1.0

florianvazelle:
### Description
Preparing for release, 3.1.0

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,8 +34,8 @@ jobs:
       - name: Run Tests
         run: |
           python -m nox -rs test-${{ matrix.entry.python-version }}
-      # - name: Upload coverage to Codecov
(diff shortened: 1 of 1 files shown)


pudo:
This would be brilliant, the current release is broken and we really want to be able to allow our customers to use AWS!

harshavamsi:
Thanks @florianvazelle, can we also bump opensearch integration testing as part of this change -- https://github.com/opensearch-project/opensearch-py/blob/main/.github/workflows/integration.yml. These are the latest versions -- https://opensearch.org/artifacts/by-version. We should bump 1.3 to 1.3.20, 2.19 to 2.19.4, and add the new 3.3.2 to the test. We should also uncomment out the codecov report here -- https://github.com/opensearch-project/opensearch-py/blob/main/.github/workflows/test.yml

PR #977, opened 2025-11-11
Add transport-grpc client library

finnegancarroll:
### Description
Adds opensearch-protobufs as a dependency.
This library is functionally unused within opensearchpy but provides clients access to protobuf schema libraries out of the box. This acts as a high level client for the transport-grpc module. Expanding opensearchpy support to this additional opensearch transport.

### Issues Resolved
N/A

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -62,7 +62,7 @@ install_requires = [
     "Events",
     # License: Apache 2.0
     # gRPC transport client libraries
-    "opensearch-protobufs==0.19.0"
+    "opensearch-protobufs==0.19.0",
 ]
 tests_require = [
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -60,6 +60,9 @@
     "python-dateutil",
     "certifi>=2024.07.04",
     "Events",
+    # License: Apache 2.0
+    # gRPC transport client libraries
+    "opensearch-protobufs==0.19.0",
```
harshavamsi: Do we need to lock this in at this version? Or can we simply point to >=?. Im wondering what happens when we bump protobufs, do we need to manually change versions here?

PR #1052, opened 2026-05-22
Fix SigV4 signature mismatch caused by hop-by-hop headers

### LINKED ISSUE #1044: [BUG] `AWSV4SignerAsyncAuth` signs `connection` header that aiohttp overrides, causing SigV4 mismatch (3.2.0 regression)
**Version:** 3.2.0 (works in 3.1.0)

**Description:**

`AWSV4Signer.sign()` in 3.2.0 now passes all request headers into `AWSRequest`, including the `connection` header set by `AsyncHttpConnection.__init__` (`connection: keep-alive`, line 161 of `http_async.py`). botocore's `SigV4Auth` then includes `connection` in `SignedHeaders`.

However, aiohttp manages the `Connection` header at the transport layer and can change it to `connection: close` after signing (e.g. when the connection pool closes a connection). AWS then computes the canonical string over `connection: close` but validates against a signature computed over `connection: keep-alive`, producing a 403 `AuthorizationException`.

**Root cause:**

The 3.2.0 change to `signer.py`:

```python
# 3.1.0
aws_request = AWSRequest(method=method.upper(), url=signature_host, data=body)

# 3.2.0
aws_request = AWSRequest(method=method.upper(), url=signature_host, data=body, headers=headers or {})
```

Pre-populating `AWSRequest` with all headers (including hop-by-hop headers) bypasses botocore's own internal filtering, which only strips headers it adds itself.

Expected behavior: Hop-by-hop headers (`connection`, `keep-alive`, `transfer-encoding`, `upgrade`, etc.) should be excluded before signing. These are transport-layer concerns that can be modified by aiohttp after the auth callback runs. botocore's SigV4Auth excludes them for this exact reason.

Suggested fix: Strip hop-by-hop headers in `http_async.py` before invoking the auth callable, or in `AWSV4Signer.sign()` before populating `AWSRequest`:
```
_HOP_BY_HOP = frozenset([
    'connection', 'keep-alive', 'proxy-authenticate', 'proxy-authorization',
    'te', 'trailers', 'transfer-encoding', 'upgrade',
])

aws_request = AWSRequest(
    method=method.upper(),
    url=signature_host,
    data=body,
    headers={k: v for k, v in (headers or {}).items() if k.lower() not in _HOP_BY_HOP},
)
```
Workaround: Subclass `AWSV4SignerAsyncAuth` and strip hop-by-hop headers in `__call__` before delegating to `super()`.

olion500:
### Description

Fixes #1044.

`AWSV4Signer.sign()` now excludes [RFC 2616 hop-by-hop headers](https://datatracker.ietf.org/doc/html/rfc2616#section-13.5.1) and any headers nominated by the `Connection` header value ([RFC 2616 Section 14.10](https://datatracker.ietf.org/doc/html/rfc2616#section-14.10)) before constructing the `AWSRequest` passed to `SigV4Auth`.

### Issue

Since v3.2.0, `AWSV4Signer.sign()` passes all request headers into `AWSRequest`, including `connection: keep-alive` set by `AsyncHttpConnection.__init__`. `SigV4Auth` then includes `connection` in `SignedHeaders`. However, `aiohttp` manages the `Connection` header at the transport layer and may modify or remove it after signing, causing a signature mismatch and HTTP 403 from AWS OpenSearch Service.

### Fix

- A `frozenset` of RFC 2616 hop-by-hop header names is used to filter headers before they reach `AWSRequest`. Only end-to-end headers are signed, preventing transport-layer modifications from invalidating the signature.
- The `Connection` header value is parsed to extract nominated headers (e.g., `Connection: keep-alive, x-custom-hop` causes `x-custom-hop` to also be excluded), per RFC 2616 Section 14.10.

### Testing

Added `TestAsyncSignerHopByHopHeaders` with two test cases:
- `test_connection_header_excluded_from_signing` — verifies `connection` is excluded from `SignedHeaders`
- `test_connection_nominated_headers_excluded_from_signing` — verifies headers nominated by the `Connection` value are also excluded

### Check List
- [x] Commits are signed with a DCO
- [x] New functionality includes testing
- [x] `CHANGELOG.md` updated (under `[Unreleased]`)
- [x] Applicable `javadocs` are added or updated — N/A (Python project)

### DIFF
--- a/opensearchpy/helpers/signer.py
+++ b/opensearchpy/helpers/signer.py
@@ -47,9 +47,39 @@ class AWSV4Signer:
 
         signature_host = self._fetch_url(url, headers or dict())
 
+        # Exclude hop-by-hop headers from signing. These headers (e.g.,
+        # "connection") may be modified or removed by the HTTP transport layer
+        # (aiohttp, urllib3) after signing, which would cause a signature
+        # mismatch and result in a 403 from AWS.
+        hop_by_hop = frozenset(
+            (
+                "connection",
+                "keep-alive",
+                "proxy-authenticate",
+                "proxy-authorization",
+                "te",
+                "trailer",
+                "transfer-encoding",
+                "upgrade",
+            )
+        )
+        headers_lower = {k.lower(): v for k, v in (headers or {}).items()}
+        conn_nominated = frozenset(
+            token.strip().lower()
+            for token in headers_lower.get("connection", "").split(",")
+            if token.strip()
+        )
+        exclude = hop_by_hop | conn_nominated
+        signing_headers = {
+            k: v for k, v in (headers or {}).items() if k.lower() not in exclude
+        }
+
         # create an AWS request object and sign it using SigV4Auth
         aws_request = AWSRequest(
-            method=method.upper(), url=signature_host, data=body, headers=headers or {}
+            method=method.upper(),
+            url=signature_host,
+            data=body,
+            headers=signing_headers,
         )
 
         # credentials objects expose access_key, secret_key and token attributes
--- a/opensearchpy/helpers/signer.py
+++ b/opensearchpy/helpers/signer.py
@@ -50,9 +50,8 @@ class AWSV4Signer:
         # Only sign headers required by the SigV4 spec: host and x-amz-*.
         # Passing all headers risks a signature mismatch when the HTTP
         # transport layer (aiohttp, urllib3) modifies them after signing.
-        headers_lower = {k.lower(): v for k, v in (headers or {}).items()}
-        signing_headers = {
-            "host": headers_lower.get("host") or urlparse(url).netloc,
+        signature_headers = {
+            "host": urlparse(signature_host).netloc,
             **{
                 k: v
                 for k, v in (headers or {}).items()
@@ -64,7 +63,7 @@ class AWSV4Signer:
             method=method.upper(),
             url=signature_host,
             data=body,
-            headers=signing_headers,
+            headers=signature_headers,
         )
 
         # credentials objects expose access_key, secret_key and token attributes
--- a/opensearchpy/helpers/signer.py
+++ b/opensearchpy/helpers/signer.py
@@ -47,34 +47,19 @@ class AWSV4Signer:
 
         signature_host = self._fetch_url(url, headers or dict())
 
-        # Exclude hop-by-hop headers from signing. These headers (e.g.,
-        # "connection") may be modified or removed by the HTTP transport layer
-        # (aiohttp, urllib3) after signing, which would cause a signature
-        # mismatch and result in a 403 from AWS.
-        hop_by_hop = frozenset(
-            (
-                "connection",
-                "keep-alive",
-                "proxy-authenticate",
-                "proxy-authorization",
-                "te",
-                "trailer",
-                "transfer-encoding",
-                "upgrade",
-            )
-        )
+        # Only sign headers required by the SigV4 spec: host and x-amz-*.
+        # Passing all headers risks a signature mismatch when the HTTP
+        # transport layer (aiohttp, urllib3) modifies them after signing.
         headers_lower = {k.lower(): v for k, v in (headers or {}).items()}
-        conn_nominated = frozenset(
-            token.strip().lower()
-            for token in headers_lower.get("connection", "").split(",")
-            if token.strip()
-        )
-        exclude = hop_by_hop | conn_nominated
         signing_headers = {
-            k: v for k, v in (headers or {}).items() if k.lower() not in exclude
+            "host": headers_lower.get("host") or urlparse(url).netloc,
+            **{
+                k: v
+                for k, v in (headers or {}).items()
+                if k.lower().startswith("x-amz-")
+            },
         }
 
-        # create an AWS request object and sign it using SigV4Auth
         aws_request = AWSRequest(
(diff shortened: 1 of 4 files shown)


olion500:
### CI Failure Analysis

The `lint`, `generate`, and `codecov/project` checks are failing, but **none of the failures are related to this PR's changes**.

All three failures trace back to a pre-existing `mypy` error on `main`:

```
test_opensearchpy/test_server/test_rest_api_spec.py:484: error: Need type annotation for "YAML_TEST_SPECS"
(hint: "YAML_TEST_SPECS: list[<type>] = ...")  [var-annotated]
```

The latest CI run on `main` ([`ffe4c7b`](https://github.com/opensearch-project/opensearch-py/commit/ffe4c7b233d61100f75edc155089f7eec2ba078c)) also fails with the same `mypy` error:
- [CI run #26261010729](https://github.com/opensearch-project/opensearch-py/actions/runs/26261010729) — `conclusion: failure`

**This PR's changes pass all relevant linting locally:** `isort`, `black`, `flake8`, and `pylint` all pass cleanly.

rbhatane (approved):

ericbn:
The last change done in AWSV4Signer was done to fix #1035, [here](https://github.com/opensearch-project/opensearch-py/commit/d8a8c573b2072ec2701819a937fb49702f928192). The issue with that change was that it is passing all headers to AWSRequest, and the opened issue itself was already hinting to what should be the right implementation:

> "You must include the host header and any x-amz-* headers in the signature."

That was not done in the last implementation, and also not being done here. The solution can be much simpler that what's being proposed here.

```python
class AWSV4Signer:
    def __init__(self, credentials, region: str, service: str = "es"):
        self._signer = SigV4Auth(credentials, service_name=service, region_name=region)

    def sign(
        self, method: str, url: str, body: Any, headers: Optional[Dict[str, str]] = None
    ) -> Dict[str, str]:
        request = AWSRequest(
            method=method,
            url=url,
            headers={
                "host": urlparse(url).netloc,
                **{k: v for k, v in headers or {} if k.lower().startswith("x-amz-")},
            },
        )
        self._signer.add_auth(request)
        return request.headers
```

opensearchpy/helpers/signer.py
```diff
@@ -47,9 +47,39 @@ def sign(
 
         signature_host = self._fetch_url(url, headers or dict())
 
+        # Exclude hop-by-hop headers from signing. These headers (e.g.,
+        # "connection") may be modified or removed by the HTTP transport layer
+        # (aiohttp, urllib3) after signing, which would cause a signature
+        # mismatch and result in a 403 from AWS.
+        hop_by_hop = frozenset(
```
harshavamsi: `accept;accept-encoding;content-length;content-type;` these are the headers we want to remove, the headers here are already black listed by boto

harshavamsi (changes requested):

opensearchpy/helpers/signer.py
harshavamsi: @olion500 can we use [this](https://github.com/opensearch-project/opensearch-py/pull/1052#issuecomment-4598126457), it is a much simpler approach

olion500:
Thanks @harshavamsi @ericbn! Updated to use the allowlist approach, only `host` and `x-amz-*` headers are signed now. Much simpler than the blocklist. 

I kept the frozen credentials logic and `_fetch_url` to handle the credential refresh race condition and cases where the host header differs from the URL's netloc.

opensearchpy/helpers/signer.py
```diff
@@ -47,9 +47,24 @@ def sign(
 
         signature_host = self._fetch_url(url, headers or dict())
 
-        # create an AWS request object and sign it using SigV4Auth
+        # Only sign headers required by the SigV4 spec: host and x-amz-*.
+        # Passing all headers risks a signature mismatch when the HTTP
+        # transport layer (aiohttp, urllib3) modifies them after signing.
+        headers_lower = {k.lower(): v for k, v in (headers or {}).items()}
+        signing_headers = {
+            "host": headers_lower.get("host") or urlparse(url).netloc,
```
ericbn: This can be simplified, since `signature_host` already contains the host name normalized. Also, I'd call the variable `signature_headers` for naming consistency.


```suggestion
        signature_headers = {
            "host": urlparse(signature_host).netloc,
```

ericbn (changes requested):

opensearchpy/helpers/signer.py
olion500: Thanks @ericbn! Applied both suggestions.
- Reuse the normalized host from `signature_host` (`urlparse(signature_host).netloc`) instead of duplicating the resolution logic.
- Renamed `signing_headers` → `signature_headers` for consistency.

ericbn:
This looks good to me in terms of the changes in the code. Looks like the tests still need to be improved though.

olion500:
> This looks good to me in terms of the changes in the code. Looks like the tests still need to be improved though.

The original test only checked the SignedHeaders substring, which tested the mechanism but not the actual bug. 

- Replaced the brittle string parsing with an assertion on the exact headers signed (`{host, x-amz-*}` only).
- Added a regression test: with the timestamp frozen, the signature is identical whether or not `Connection/Content-Type` are present.
- Added a `requests`-path test that signs a real `PreparedRequest`, reproducing the failure with the headers `requests` actually injects.

PR #1053, opened 2026-05-22
Pin GitHub Actions to commit SHAs

Divyaasm:
### Description

Pin all GitHub Action tag references to their corresponding commit SHAs.

Tags are mutable references that can be force-pushed to point to different commits, making them vulnerable to supply chain attacks. Commit SHAs are immutable and guarantee that the exact reviewed code is executed in CI workflows. This change pins all third-party actions to their current commit SHAs to prevent potential tampering.

### DIFF
--- a/.github/workflows/backport.yml
+++ b/.github/workflows/backport.yml
@@ -15,14 +15,14 @@ jobs:
     steps:
       - name: GitHub App token
         id: github_app_token
-        uses: tibdex/github-app-token@v2.1.0
+        uses: tibdex/github-app-token@3beb63f4bd073e61482598c45c71c1019b59b73a # v2.1.0
         with:
           app_id: ${{ secrets.APP_ID }}
           private_key: ${{ secrets.APP_PRIVATE_KEY }}
           installation_id: 22958780
 
       - name: Backport
-        uses: VachaShah/backport@v2.2.0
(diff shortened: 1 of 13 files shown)


jayaddison:
NB: although pinning GitHub Actions versions to SHA references is a commonly-accepted good practice, it has a side-effect of disabling GitHub Dependabot Alerts.

Roughly speaking, this is because Dependabot Alerts detects vulnerabilities based on semver ranges, and it isn't (yet?) aware of the version range a SHA commit sits within (if any).

For more info, see:
 - https://docs.github.com/en/code-security/concepts/supply-chain-security/about-dependabot-alerts#limitations
 - https://github.com/orgs/community/discussions/154189
 - https://github.com/orgs/zizmorcore/discussions/1898

Alternative vulnerability scanners that can detect vulnerable GitHub Actions, including cases where those actions are pinned by SHA commit, do exist.

harshavamsi (approved):

finnegancarroll:
Hi @Divyaasm thanks for raising this! It looks like backport is failing due to some GAs which are not pinned.

Other checks are failing due to unrelated annotation issue. @harshavamsi  maybe this can be force merged when the above is resolved?
```
test_opensearchpy/test_server/test_rest_api_spec.py:484: error: Need type annotation for "YAML_TEST_SPECS" (hint: "YAML_TEST_SPECS: list[<type>] = ...")  [var-annotated]
```

PR #1057, opened 2026-06-04
Adding gRPC python low level client - translation layer, grpc transport, and integration tests

dayana-j:
**### Description**
This PR is for adding translation layer that converts the python low level client code to protobuf and back to the python client in the same representation they sent their data in under the hood into the main branch. There are several integration tests that was used to make sure the code was running properly. There is also gRPC transport class added to help with the transport for OpenSearch-py using gRPC, bulk request script and unit tests for it. 


opensearch_grpc/proto/common_pb2.py
```diff
@@ -0,0 +1,638 @@
+# -*- coding: utf-8 -*-
```
finnegancarroll: These compiled protobuf files should come in with the `opensearch-protobufs` optional dependency added in setup.py so we shouldn't need to include them manually here.

opensearch_grpc/grpc_transport.py
```diff
@@ -0,0 +1,262 @@
...
+        handler = self._get_grpc_handler(method, url)
+        if handler:
+            return handler(method, url, params, body)
+
+        return super().perform_request(
+            method, url, params=params, body=body,
+            timeout=timeout, ignore=ignore, headers=headers
+        )
+
+    def _get_grpc_handler(self, method, url):
```
finnegancarroll: I think this is a great place to hook in and handle gRPC since it lets us ignore whatever high level methods and objects are provided by the python client.

opensearch_grpc/grpc_transport.py
```diff
@@ -0,0 +1,262 @@
...
+        routing = params.get("routing") if params else None
+
+        converter = RequestConverter.from_body(
+            body, index=url_index, refresh=refresh,
+            timeout=timeout, pipeline=pipeline, routing=routing,
+        )
+        response = self._document_stub.Bulk(converter.build())
+        return ResponseConverter._convert_bulk_items(response)
+
+    def _handle_search(self, method, url, params, body):
```
finnegancarroll: Do we need these gRPC handlers for methods we are not handling in this PR? I think we should already handle falling back to REST search here:
https://github.com/opensearch-project/opensearch-py/pull/1057/changes#diff-bedcb616ebeb0c2214942cf7115d4e14c1c4cb39bd9313a725e01ce076473450R99-R102

opensearch_grpc/proto_adapter.py
```diff
@@ -0,0 +1,48 @@
...
+# When opensearch-protobufs publishes real modules, replace these two lines:
+#   from opensearch_protobufs.schemas import common_pb2
+#   from opensearch_protobufs.services import document_service_pb2_grpc
+
+from opensearch_grpc.proto import common_pb2, document_service_pb2_grpc
+
+# ─── Re-export everything that other modules need ─────────────────────────────
+
+# Protobuf message types (for building requests)
+BulkRequest = common_pb2.BulkRequest
```
finnegancarroll: Not sure we need this file. The names of these `opensearch-protobufs` objects should not change.

test_opensearchpy/test_translation/test_bulk_stream.py
```diff
@@ -0,0 +1,85 @@
...
+        {"create": {"_index": "grpc-test", "_id": "3"}},
+        {"title": "Third doc", "value": 3},
+        {"update": {"_index": "grpc-test", "_id": "1"}},
+        {"doc": {"value": 100}},
+        {"delete": {"_index": "grpc-test", "_id": "2"}},
+    ]
+    return toProtoBulkRequest(body=body, refresh="true")
+
+
+def test_bulk_stream():
```
finnegancarroll, line 50: Lets remove these files for manually testing against a local node until we can run them as integration tests.

setup.py
```diff
@@ -70,6 +70,7 @@
     "pytz",
     "botocore",
     "pytest-mock<4.0.0",
+    "opensearch-protobufs==1.4.0",
```
finnegancarroll, line 73: opensearch-protobufs should stay as an optional dependency only. We can guard our usage of gRPC in the client by checking inline if we are able to import the needed dependencies.

```
try:
    import grpcio
```

.gitignore
```diff
@@ -153,4 +153,14 @@ test_opensearch/local.py
 *~
 
 # Visual Studio Code
-.vscode
\ No newline at end of file
+.vscode
+
+# Local scratch files (not for PR)
```
finnegancarroll, line 158: Lets remove these specific test files from .gitignore. Usually we only want to add very broad and common files here which may be created by common tooling/ide.

opensearch_grpc/stream_client.py
```diff
@@ -0,0 +1,225 @@
...
+    - https://docs.opensearch.org/latest/clients/python-low-level/
+"""
+
+import grpc
+
+from opensearch_grpc.proto_adapter import DocumentServiceStub
+from opensearch_grpc.translation import RequestConverter, ResponseConverter
+
+
+class StreamClient:
```
finnegancarroll: I'm not sure we need this new client behavior to manage our gRPC requests. We don't implement any actual streaming endpoints in gRPC today and the underlying grpc.io library will handle any queueing behavior for us.

Instead here we should be providing an `OpenSearchGrpc` client which extends the existing `OpenSearch` client and this new `class GrpcTransport(Transport):` to route requests to gRPC or else fallback to REST.

For the initial PR it may be best to separate out the translation logic for converting python client Bulk requests to their protobuf representations, and the client code which handles the underlying gRPC connection/channel.

opensearch_grpc/grpc_transport.py
```diff
@@ -0,0 +1,262 @@
...
+
+    def __init__(self, hosts, *args, **kwargs):
+        self._grpc_port = kwargs.pop("grpc_port", 9400)
+        self._grpc_host_override = kwargs.pop("grpc_host", None)
+        self._grpc_hosts = kwargs.pop("grpc_hosts", None)
+
+        super().__init__(hosts, *args, **kwargs)
+
+        # Resolve gRPC target from grpc_hosts, grpc_host, or hosts
+        if self._grpc_hosts:
```
finnegancarroll: nit: i think it might be best to be explicit when a user configures a gRPC client. Falling back to the rest host/port or localhost could be un-intuitive and lead to unexpected errors.

opensearch_grpc/grpc_transport.py
```diff
@@ -0,0 +1,262 @@
+"""
+grpc_transport.py — gRPC Transport for the opensearch-py Client
+
+Routes operations over gRPC where supported by the OpenSearch server:
+    - bulk   → DocumentService.Bulk (native gRPC)
```
finnegancarroll: Initially i think we only need to handle bulk here. Other gRPC handlers can be added incrementally and for some, such as create/delete i'm not sure there's much benefit in translating these to single item bulk requests.

The main operation we want to capture here is bulk requests with multiple documents. Something like index creation (which is pretty infrequent and has a small payload) is unlikely to benefit much from gRPC.

finnegancarroll (changes requested):

opensearch_grpc/proto/common_pb2.py
dayana-j: Thank you for pointing that out. I made the changes.

opensearch_grpc/proto_adapter.py
dayana-j: I agree since the compiling of protobuf will be handled using the opensearch-protobufs in setup.py

PR #1060, opened 2026-06-15
Remove dependency on Events package

### LINKED ISSUE #756: [BUG] Events import failed after updating to v2.6
### What is the bug?
I have a project with a folder named `events` and after updating the opensearch-py from 2.5 to 2.6, where the `service time metrics` was added, I'm receiving the error below because python tries to import the `Events` from my project's home directory and not from the Events lib.
As `events` is a common name for folders and it might cause some trouble with other people too, do you have any suggestion for fixing this issue without renaming my project's folder to another name?

```text
 File "/home/[redacted]/venv/lib/python3.11/site-packages/opensearchpy/metrics/metrics_events.py", line 13, in <module>
    from events import Events
ImportError: cannot import name 'Events' from 'events'
```

Add a folder `events` to your project and them call `from opensearchpy import OpenSearch`

python3.11 + ubuntu 20 + opensearch 2.6

noelleleigh:
### Description

Remove dependency on [Events](https://pypi.org/project/Events/) package to eliminate frequent module name collisions when `opensearch-py` is integrated into existing codebases which already have a module named `events`.

The usage of Events from #716 amounted to calling two functions in `opensearchpy.metrics.metrics_events.MetricsEvents`, so removing the indirection from Events was simple. This change has no effect on the public or private API of `opensearchpy.metrics.metrics_events.MetricsEvents`.

### Issues Resolved

Fixes #756

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Security
 ### Dependencies
 - Refresh `benchmarks/` poetry lock to pick up aiohttp 3.13.5, urllib3 2.6.3, requests 2.33.1, pygments 2.20.0 (clears Mend CVE findings) ([#1046](https://github.com/opensearch-project/opensearch-py/pull/1046))
+- Removed dependency on [Events](https://pypi.org/project/Events/)
 
 ## [3.2.0]
 ### Added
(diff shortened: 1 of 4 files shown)


noelleleigh:
> ⚠️ Report is 139 commits behind head on main.

I think the coverage regression is a false positive due to this issue.

harshavamsi (approved):
Thanks!

PR #1077, opened 2026-06-24
fix: don't short-circuit bytes in JSONSerializer.dumps (#488)

### LINKED ISSUE #488: [BUG] Opensearch Bulk API doesn't support bytes even if Serializer typing says it does
### What is the bug?
I wanted to use an alternative serializer to speed up performance. Looking at the original JsonSerializer, if the type is a [`string_types` which is either str or bytes](https://github.com/opensearch-project/opensearch-py/blob/c8b04a5246dad6dcec05cf6f5bbf2eb14a9ec398/opensearchpy/compat.py#L48), it will return the data without using the serializer. So assumably any serializer that returns str or bytes is also valid. However, the following serializer using orjson returns bytes and [bulk API only accepts strings](https://github.com/opensearch-project/opensearch-py/blob/c8b04a5246dad6dcec05cf6f5bbf2eb14a9ec398/opensearchpy/helpers/actions.py#L117C1-L122C51) as it tries and encode the data using utf8. The [inbuilt JSONSerializer ](https://github.com/opensearch-project/opensearch-py/blob/c8b04a5246dad6dcec05cf6f5bbf2eb14a9ec398/opensearchpy/serializer.py#L151C1-L153) says it takes string_types as well.
```python
class OrjsonSerializer(JSONSerializer):
    def dumps(self, data: Any):
        if isinstance(data, string_types):
            return data

        try:
        	# The .decode("utf-8") was added to convert bytes to string
            return orjson.dumps(data, default=self.default).decode("utf-8") 
        except Exception as exc:
            raise SerializationError(data, exc) from exc

    def loads(self, s: Any):
        try:
            return orjson.loads(s)
        except Exception as exc:
            raise SerializationError(s, exc) from exc
```

Return bytes from the Json Serializer instead of a string.

I'm not sure what the correct answer is but it should either only accepts strings or have a separate code path if the type is bytes

Debian 5.10.191-1

image

_Add any other context about the problem._

PraveenKumarVk:
### Description
JSONSerializer.dumps() short-circuits any input matching isinstance(data, string_types),
where string_types = (str, bytes). Because bytes is included, raw bytes passed to dumps()
are returned unchanged instead of going through serialization. This silently breaks custom
serializer subclasses (e.g. ones using orjson) that depend on the fallback branch to decode
bytes to str before the data reaches the bulk API, which expects strings only.

This change narrows the check to isinstance(data, str), so bytes no longer match the
short-circuit and instead fall through to the normal serialization path. As a result,
passing raw bytes to dumps() now raises SerializationError instead of silently passing
corrupted data downstream.

The shared string_types constant in compat.py is left unchanged, since it's used
elsewhere (client/utils.py, connection/http_requests.py, connection/http_async.py,
helpers/actions.py) for unrelated checks outside the scope of this fix.

Added tests confirming:
- dumps() still returns an already-serialized str unchanged
- dumps() still correctly serializes a dict via json.dumps
- dumps() no longer short-circuits bytes, and raises SerializationError instead

### Issues Resolved
Closes #488

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 ### Removed
 ### Fixed
 - Fix `UnicodeEncodeError` on surrogate/emoji characters in bulk chunk sizing by passing `surrogatepass` error handler ([#1086](https://github.com/opensearch-project/opensearch-py/pull/1086))
+- Fixed `_ActionChunker.feed()` not handling bytes returned from a custom serializer's `dumps()`, and reverted an unrelated regression in `JSONSerializer.dumps()` bytes handling ([#1077](https://github.com/opensearch-project/opensearch-py/pull/1077))
 ### Security
 - Refresh `samples/` and `benchmarks/` poetry locks and pin fixed transitive floors to clear all outstanding dependency CVE findings: aiohttp `>=3.14.1` (CVE-2026-54273..54280 et al.), urllib3 `>=2.7.0` (CVE-2025-50181/-50182/-66418/-66471, CVE-2026-21441/-44431/-44432), requests `>=2.33.0` (CVE-2024-47081, CVE-2026-25645), and idna `>=3.18` (CVE-2026-45409) ([#1082](https://github.com/opensearch-project/opensearch-py/pull/1082))
 ### Dependencies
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -1,32 +1,56 @@
 # CHANGELOG
+
```
harshavamsi: Hi, looks like the changelog has changes that are not required -- new lines, extra entry etc. Could you fix this, thanks!

harshavamsi (changes requested):

PR #1078, opened 2026-06-24
gRPC Transport/Connection Management 

dayana-j:
### Description
This PR introduces GrpcTransport, a transport class that extends the standard opensearch-py Transport, OpenSearchGrpc client class, and gRPC transport plugins for docker. When a bulk request is made, GrpcTransport intercepts it and routes the operation over gRPC using the BulkRequestProtoBuilder to translate the client's NDJSON-style bulk body into protobuf format. The ResponseConverter then translates the protobuf response back into the dictionary format that existing opensearch-py users expect — including took, errors, items, status codes, and shard information. For any non-bulk operation (search, get, count, index management), the transport falls through to REST automatically with no additional logic required.
```
class GrpcTransport(Transport):
  def perform_request(self, method, url, params=None, body=None, timeout=None, ignore=(), headers=None):
      """Route to gRPC or REST based on the URL pattern."""
      # Check if this request can be handled via gRPC
      handler = self._get_grpc_handler(method, url)
      if handler:
          return handler(method, url, params, body)

      # Fall back to REST for all non-bulk operations
      return super().perform_request(method, url, params=params, body=body, timeout=timeout, ignore=ignore,
      headers=headers)

  def _handle_bulk(self, method, url, params, body):
      """Bulk → DocumentService.Bulk (native gRPC)."""
      # Extract index from URL (e.g. /my-index/_bulk → 'my-index')
      url_index = self._extract_index_from_url(url, "_bulk")

      # Extract query parameters passed by opensearch-py
      refresh = params.get("refresh") if params else None
      timeout = params.get("timeout") if params else None
      pipeline = params.get("pipeline") if params else None
      routing = params.get("routing") if params else None

      # Translate opensearch-py bulk body (NDJSON pairs) into protobuf BulkRequest
      converter = BulkRequestProtoBuilder.from_body(
          body, index=url_index, refresh=refresh, timeout=timeout, pipeline=pipeline, routing=routing)

      # Send over gRPC and convert protobuf response back to opensearch-py format
      response = self._document_stub.Bulk(converter.build())
      return ResponseConverter._convert_bulk_items(response)
```

OpenSearchGrpc, a client class that extends the standard OpenSearch client. It uses GrpcTransport as its transport class, which routes bulk requests over gRPC while all other operations fall back to REST automatically. The standard OpenSearch client is unchanged — existing customers are not affected.
```
class OpenSearchGrpc(OpenSearch):
      def __init__(self, hosts=None, grpc_hosts=None, **kwargs):
          from opensearch_grpc.grpc_transport import GrpcTransport

          kwargs["grpc_hosts"] = grpc_hosts
          super().__init__(hosts, transport_class=GrpcTransport, **kwargs)
```

Users who want gRPC simply import OpenSearchGrpc instead of OpenSearch and add the grpc_hosts parameter. All their existing parameters (auth, SSL, etc.) continue to work as before.

### Usage Example
```
  from opensearchpy import OpenSearchGrpc

  client = OpenSearchGrpc(
      hosts=[{'host': 'localhost', 'port': 9200}],
      grpc_hosts=[{'host': 'localhost', 'port': 9400}],
  )

  # Bulk goes over gRPC automatically
  resp = client.bulk(body=[
      {"index": {"_index": "my-index", "_id": "1"}},
      {"title": "Document 1", "content": "Hello"},
  ], refresh=True)

  # Everything else uses REST — no changes needed
  results = client.search(index="my-index", body={"query": {"match_all": {}}})
```

### Connection Management and CI Infrastructure

To enable end-to-end gRPC communication, the OpenSearch Docker image needed to be configured with the gRPC transport plugin. The .ci/opensearch/Dockerfile now adds the necessary settings to opensearch.yml: 

```
ARG OPENSEARCH_VERSION
  RUN if echo "$OPENSEARCH_VERSION" | grep -qE "^3\."; then \
        echo "aux.transport.types: [transport-grpc]" >> $opensearch_yml \
        && echo "aux.transport.transport-grpc.port: 9400" >> $opensearch_yml \
        && echo "grpc.host: 0.0.0.0" >> $opensearch_yml; \
      fi

```
aux.transport.types:[transport-grpc] enables the gRPC transport plugin, aux.transport.transport-grpc.port: 9400 configures the port. OpenSearch listens on for gRPC connections, and grpc.host: 0.0.0.0 binds the gRPC listener to all interfaces so the client container can reach it. These settings are applied conditionally — only when the OpenSearch version is 3.x — because the gRPC transport plugin does not exist in OpenSearch 1.x or 2.x. This ensures existing CI test runs against older versions continue to work without modification. The run-opensearch.sh script publishes port 9400 alongside the existing port 9200 in the docker run command, establishing the full network path for gRPC traffic from the test client container to the OpenSearch server.

### Integration Tests

This PR also introduces several integration tests.  The test_opensearchpy/test_server/test_grpc/ directory contains a comprehensive suite of integration tests that validate the entire gRPC pipeline from client to server and back. All tests inherit from OpenSearchGrpcTestCase, which performs a module-level gRPC availability check and skips gracefully when the gRPC port is unreachable (e.g.,OpenSearch 2.x), making them safe for CI environments that don't have gRPC available.

The core validation is the large bulk upload test, which uploads 10 documents in a single gRPC bulk request and verifies the response format matches what the standard Python client returns: 
```
def test_upload_documents(self, client, index_name):
      body = []
      for i in range(10):
          body.append({"index": {"_index": index_name, "_id": str(i)}})
          body.append({"title": f"Document {i}", "category": f"category-{i % 10}", "value": i})

      resp = client.bulk(body=body, refresh=True)

      assert resp["errors"] is False
      assert len(resp["items"]) == 10
      assert resp["items"][0]["index"]["status"] == 201
      assert resp["items"][0]["index"]["result"] == "created"
```
It validates every response field (_index, _id, result, status, _version, _shards), then performs follow-up operations: updating 5 documents, deleting 2, and executing a mixed batch of index + update + delete in one call. The same test is duplicated using the OpenSearchGrpc client to prove both test paths work identically.

Additional tests cover index, create, update, delete with proper error handling for conflicts and 404s, and REST fallback verification (search, get, count, and index management all pass through correctly).
 

opensearch_grpc/grpc_transport.py
```diff
@@ -0,0 +1,148 @@
...
+    """
+    Transport that routes bulk operations over gRPC.
+
+    Bulk requests are sent via DocumentService.Bulk for better performance.
+    All other operations fall back to REST automatically.
+    """
+
+    # Operations we handle via gRPC (URL patterns)
+    _GRPC_ROUTES = {
+        "_bulk": "_handle_bulk",
```
finnegancarroll: is `_handle_bulk` a valid OpenSearch API?

finnegancarroll: we might want to match on a regex pattern because we can invoke Bulk API in a couple different ways:
https://docs.opensearch.org/latest/api-reference/document-apis/bulk/#endpoints

Users can have an index name which happens to contain "_bulk" as well.

opensearch_grpc/grpc_transport.py
```diff
@@ -0,0 +1,148 @@
...
+        self._grpc_port = kwargs.pop("grpc_port", 9400)
+        self._grpc_host_override = kwargs.pop("grpc_host", None)
+        self._grpc_hosts = kwargs.pop("grpc_hosts", None)
+
+        super().__init__(hosts, *args, **kwargs)
+
+        # Resolve gRPC target from grpc_hosts, grpc_host, or hosts
+        if self._grpc_hosts:
+            # Use grpc_hosts parameter: [{"host": "x", "port": 9400}]
+            first_grpc = (
```
finnegancarroll: I think there's some functionality in the python client where you can provide multiple endpoints and it will perform some "round robin" sending requests to each. I think we don't need to handle this for our first version. Lets just throw an exception and give the user an error if they provide multiple gRPC hosts here.

~"Multiple gRPC host targets not yet supported"

opensearch_grpc/grpc_transport.py
```diff
@@ -0,0 +1,148 @@
...
+                self._grpc_hosts[0]
+                if isinstance(self._grpc_hosts[0], dict)
+                else {"host": self._grpc_hosts[0]}
+            )
+            grpc_host = first_grpc.get("host", "localhost")
+            grpc_port = first_grpc.get("port", self._grpc_port)
+        elif self._grpc_host_override:
```
finnegancarroll: I think we can drop these fallback cases. If a user does not supply a gRPC and host parameters we should not be making any changes to their client behavior.

opensearch_grpc/grpc_transport.py
```diff
@@ -0,0 +1,148 @@
...
+            grpc_host = "localhost"
+            grpc_port = self._grpc_port
+
+        self._grpc_address = f"{grpc_host}:{grpc_port}"
+        self._channel = grpc.insecure_channel(self._grpc_address)
+        self._document_stub = document_service_pb2_grpc.DocumentServiceStub(
+            self._channel
+        )
+        self._search_stub = search_service_pb2_grpc.SearchServiceStub(self._channel)
+        self._serializer = JSONSerializer()
```
finnegancarroll: should this already be handled by the parent class constructor?

opensearch_grpc/grpc_transport.py
```diff
@@ -0,0 +1,148 @@
...
+        else:
+            grpc_host = "localhost"
+            grpc_port = self._grpc_port
+
+        self._grpc_address = f"{grpc_host}:{grpc_port}"
+        self._channel = grpc.insecure_channel(self._grpc_address)
+        self._document_stub = document_service_pb2_grpc.DocumentServiceStub(
+            self._channel
+        )
+        self._search_stub = search_service_pb2_grpc.SearchServiceStub(self._channel)
```
finnegancarroll: we can remove this since we aren't supporting search in this version.

opensearchpy/client/__init__.py
```diff
@@ -194,6 +194,8 @@ def __init__(
         self,
         hosts: Any = None,
         transport_class: Type[Transport] = Transport,
+        grpc: bool = False,
```
finnegancarroll: i think this is redundant with grpc_hosts. if a user supplies grpc_hosts it implies grpc=true.

finnegancarroll: im thinking in general it will probably be better to use inheritance here. provide a GrpcOpenSearchClient which inherits from OpenSearchClient.

This design might make a little more sense because we will not be adding inert code to OpenSearchClient which will be completely non functional for everyone not using gRPC. It also reduces the risk of accidentally breaking the baseline functionality of the OpenSearchClient.

opensearchpy/client/__init__.py
```diff
@@ -3274,3 +3290,46 @@ def bulk_stream(
             headers=headers,
             body=body,
         )
+
+
+
+class OpenSearchGrpc(OpenSearch):
```
finnegancarroll: oh got it i see we do have a separate client class for gRPC. We should move all of our gRPC specific client behavior into this class.

test_opensearchpy/test_translation/test_bulk_large_upload.py
```diff
@@ -0,0 +1,271 @@
...
+try:
+    _channel = grpc.insecure_channel(f"localhost:{_grpc_port}")
+    grpc.channel_ready_future(_channel).result(timeout=2)
+    _channel.close()
+except Exception:
+    pytest.skip("gRPC server not available", allow_module_level=True)
+
+
+@pytest.fixture(scope="module")
+def client():
```
finnegancarroll: i think we should be able to leverage the existing integ test framework client creation instead of re-implementing here.

test_opensearchpy/test_translation/test_bulk_large_upload.py
```diff
@@ -0,0 +1,271 @@
...
+
+@pytest.fixture(autouse=True, scope="module")
+def cleanup(client, index_name):
+    client.indices.delete(index=index_name, ignore=[404])
+    yield
+    client.indices.delete(index=index_name, ignore=[404])
+
+
+class TestLargeBulkUpload:
+    """Simulate a real client uploading 10K+ documents over gRPC."""
```
finnegancarroll: it looks like we removed the 10k+ documents test case. May want to rename this class and comment to reflect that.

test_opensearchpy/test_translation/test_bulk_large_upload.py
```diff
@@ -0,0 +1,271 @@
...
+
+    def test_verify_document_count(self, client, index_name):
+        """Verify all 10 documents are searchable."""
+        print("\n[TEST] Verifying document count...")
+        client.indices.refresh(index=index_name)
+        resp = client.count(index=index_name)
+        count = resp["count"]
+        print(f"[TEST] Document count: {count}")
+        assert count == 10, f"Expected 10, got {count}"
+        print("[TEST] ✅ All 10 documents confirmed in index")
```
finnegancarroll: these tests LGTM. we can remove these test print outputs for the final version.

CHANGELOG.md
```diff
@@ -3,6 +3,11 @@ Inspired from [Keep a Changelog](https://keepachangelog.com/en/1.0.0/)
 
 ## [Unreleased]
 ### Added
+- Add `OpenSearchGrpc` client class extending `OpenSearch` with built-in gRPC transport for bulk operations
```
finnegancarroll: typically a changelog entry corresponds to a single PR. can we summarize these changes into one entry? for detailed breakdown of the changes we provide a link to the associated PR for further reading.

https://github.com/opensearch-project/opensearch-py/blob/main/CHANGELOG.md

finnegancarroll (changes requested):

opensearch_grpc/grpc_transport.py
dayana-j: Yes it is. Thanks for pointing this out.

.ci/opensearch/Dockerfile
```diff
@@ -6,3 +6,10 @@ ARG opensearch_yml=$opensearch_path/config/opensearch.yml
 
 ARG SECURE_INTEGRATION
 RUN if [ "$SECURE_INTEGRATION" != "true" ] ; then echo "plugins.security.disabled: true" >> $opensearch_yml; fi
+
+ARG OPENSEARCH_VERSION
+RUN if echo "$OPENSEARCH_VERSION" | grep -qE "^[3-9]\."; then \
```
harshavamsi: So this is saying that any OpenSearch version starting from 3-9? Why dont we pin it after the version that we have gRPC?

dayana-j: Yes it is. But I can pin it after the version 3.x that we have gRPC in the event of a security risk arises from a future version 4.x and beyond will not affect the client. It will have to be updated in the future for any version beyond the last 3.x version. I will make that change.

opensearchpy/client/__init__.py
```diff
@@ -3274,3 +3274,45 @@ def bulk_stream(
...
+
+        # Everything else uses REST
+        client.search(index='my-index', body={'query': {'match_all': {}}})
+
+    :arg hosts: list of REST nodes (same as OpenSearch client).
+    :arg grpc_hosts: list of gRPC nodes, e.g. [{'host': 'localhost', 'port': 9400}].
+    :arg kwargs: all other arguments passed to the OpenSearch client.
+    """
+
+    def __init__(
```
finnegancarroll: How are we handling features/args which are supported in the REST client but not supported on gRPC? For example TLS. I think here we need to be very explicit, if someone tries to create an `OpenSearchGrpc` with any TLS arguments, we should fail to create the client and throw some exception explaining this is not yet supported in the gRPC client.

opensearch_grpc/grpc_transport.py
```diff
@@ -0,0 +1,145 @@
...
+
+        converter = BulkRequestProtoBuilder.from_body(
+            body,
+            index=url_index,
+            refresh=refresh,
+            timeout=timeout,
+            pipeline=pipeline,
+            routing=routing,
+        )
+        response = self._document_stub.Bulk(converter.build())
```
finnegancarroll: Do we know how errors are typically handled in the python client? I think it may be worth looking into this for a couple reasons.

1. If the REST client typically throws some `pensearchpy.exceptions.ConnectionError` exception type, we should try to mimic this in our gRPC client. That way if someone is catching this specific exception to handle connection errors, the behavior doesn't change.

2. There might be some retry or health check functionality baked into the underlying REST transport which we want to mimic here. I think these are probably less important and may already be handled by the underlying gRPC libraries we are using but worth checking to see or adding a "TODO" for these features.

opensearchpy/client/__init__.py
```diff
@@ -3274,3 +3274,45 @@ def bulk_stream(
...
+        self,
+        hosts: Any = None,
+        grpc_hosts: Any = None,
+        **kwargs: Any,
+    ) -> None:
+        from opensearch_grpc.grpc_transport import GrpcTransport
```
finnegancarroll: Im wondering what the user sees when they try to create a gRPC client, but have not installed the optional gRPC dependencies? Depending on what this error message looks like we may want to add a try/catch here to check if these gRPC dependencies are available and provide a clear exception if they are not.

opensearchpy/client/__init__.py
finnegancarroll: Same with auth params as well. If we see any auth args passed in this version we need to throw an exception since gRPC doesn't support these. It may be worth documenting in code the possible constructor args here for an OpenSearch client and specifying which are applied to gRPC client as well, which are unsupported and throw errors, and which gRPC can safely ignore.

finnegancarroll (changes requested):

opensearchpy/client/__init__.py
dayana-j: The user would get an import error saying No module named 'opensearch.protobufs' or 'grpc'. I have added in a try/catch to provide a clear exception using the Improperly configured exception from the exception.py file in the codebase. They will get this message if they do not have the grpc dependencies installed: ImproperlyConfigured: gRPC dependencies are not installed. Install them with: pip install opensearch-py[grpc]

test_opensearchpy/test_server/test_grpc/test_bulk.py
```diff
@@ -0,0 +1,174 @@
...
+test_bulk.py — gRPC Bulk Integration Tests
+
+Tests bulk operations routed over gRPC using the standard OpenSearch client
+with grpc=True. Validates response format, mixed operations, and error handling.
+"""
+
+from . import OpenSearchGrpcTestCase
+
+
+class TestBulkIndex(OpenSearchGrpcTestCase):
```
finnegancarroll, line 22: Can we consolidate these tests? It doesn't seem like there is a functional difference between test_bulk, test_bulk_request, test_bulk_stream, ect. Maybe we can narrow these down into only a few key files.

I think test_bulk_large_upload is similarly unnecessary. We don't need to stress test here and should only focus on correctness.

opensearch_grpc/grpc_transport.py
```diff
@@ -0,0 +1,205 @@
...
+            POST /<index>/_bulk
+        """
+        if method in ("POST", "PUT") and self._BULK_PATTERN.match(url):
+            return self._handle_bulk
+
+        return None
+
+    # ─── gRPC Handlers ────────────────────────────────────────────────────────
+
+    def _handle_bulk(
```
finnegancarroll, line 158: Are we missing `require_alias` param here?

opensearch_grpc/grpc_transport.py
```diff
@@ -0,0 +1,205 @@
...
+        )
+
+        try:
+            response = self._document_stub.Bulk(converter.build())
+        except grpc.RpcError as e:
+            self._raise_grpc_error(e)
+
+        return ResponseConverter._convert_bulk_items(response)
+
+    def _raise_grpc_error(self, error: grpc.RpcError) -> None:
```
finnegancarroll, line 186: Lets include the full comprehensive mappings here just to ease future development and automatically handle any unexpected errors thrown:
https://github.com/opensearch-project/OpenSearch/issues/18926

finnegancarroll (changes requested):

opensearch_grpc/grpc_transport.py
dayana-j: Okay

finnegancarroll (approved):

finnegancarroll:
LGTM. I see the failing integ test is api spec related and probably flaky. Only other thing is code cov. It looks like it's comparing against a much older commit but can you check this and see if we have sufficient coverage for these changes?

opensearchpy/client/__init__.py
```diff
@@ -37,6 +37,7 @@
 import logging
 from typing import Any, Type
 
+from ..exceptions import ImproperlyConfigured
```
harshavamsi: A high level comment here, the OpenSearch Python client is supposed to be generated from API Spec, you can find the generator code here -- https://github.com/opensearch-project/opensearch-py/blob/main/utils/generate_api.py. We should be thinking about how we can handle this in the generator. I understand for the purposes of the tests we need this, but let's do some follow up work to get this into the generator

opensearchpy/client/__init__.py
```diff
@@ -3274,3 +3275,98 @@ def bulk_stream(
             headers=headers,
             body=body,
         )
+
+
+class OpenSearchGrpc(OpenSearch):
```
harshavamsi: Trying to understand the design decisions here. Why does this need to be a new client? We could still initialize the original client as 

```
       client = OpenSearch(
            hosts=[{'host': 'localhost', 'port': 9200}],
            grpc=True,
            grpc_hosts=[{'host': 'localhost', 'port': 9400}],
        )
```

We can abstract away the grpc logic within the grpc transport above

opensearch_grpc/grpc_transport.py
```diff
@@ -0,0 +1,246 @@
...
+                except TransportError as e:
+                    if (
+                        hasattr(e, "status_code")
+                        and e.status_code in self.retry_on_status
+                        and attempt < self.max_retries
+                    ):
+                        continue
+                    raise
+
+        return super().perform_request(
```
harshavamsi, line 128: I commented on another PR, but see https://github.com/opensearch-project/opensearch-py/pull/1087#discussion_r3555853319

opensearchpy/client/__init__.py
dayana-j: Okay I will do that.

opensearchpy/client/__init__.py
dayana-j: A grpc client was made to help with distinguishing between the standard client and grpc. We did not want the another clients that is already using the standard OpenSearch client to accidentally initiate grpc when they do not intend to.

opensearch_grpc/grpc_transport.py
dayana-j: GrpcTranport extends the default transport, so we are not missing the client arguments that the default transport supports. The REST fallback would inherit all of the client arguments because gRPC transport extending the default transport.

finnegancarroll (approved):

finnegancarroll:
LGTM

harshavamsi (approved):

PR #1087, opened 2026-07-06
TLS/mTLS, Basic Authentication, and JWT support for gRPC transport

dayana-j:
### Description
This PR adds TLS, mutual TLS (mTLS), Basic authentication, and JWT (JSON Web Token) authentication support to the gRPC transport, enabling secure and authenticated communication between the Python client and OpenSearch's gRPC endpoint. The implementation reuses the same parameters that existing REST clients already use (use_ssl, ca_certs, client_cert, client_key, http_auth), so users familiar with the standard OpenSearch client can enable security on gRPC without learning new configuration options. JWT implementation allows users to pass a pre-obtained JWT token via the http_auth parameter, and the token is attached as a Bearer authorization header in gRPC metadata on every call. It follows the same interceptor pattern used for Basic auth — a BearerTokenInterceptor that injects the token into gRPC metadata. This PR also adds in modification to the transport and translation layer to help remove the extra overhead that was observed when benchmarking was done on the code using profiling.

### TLS/mTLS
When use_ssl=True is passed to OpenSearchGrpc, the transport creates a grpc.secure_channel instead of a grpc.insecure_channel. The default behavior remains use_ssl=False (insecure channel). Existing users without TLS are completely unaffected.

The channel credentials are built from the same certificate parameters the REST client accepts. If ca_certs is provided, the file is read and passed as the root certificate authority for server verification. If no ca_certs is specified, gRPC uses the system's default trusted CAs. For mutual TLS, providing client_cert and client_key loads the client certificate chain and private key into the channel credentials, allowing the server to verify the client's identity.
```
   # Create channel — secure (TLS/mTLS) or insecure
   if self._use_ssl:
       root_certs = None
       if self._ca_certs:
           with open(self._ca_certs, "rb") as f:
               root_certs = f.read()

       private_key = None
       cert_chain = None
       if self._client_cert:
           with open(self._client_cert, "rb") as f:
               cert_chain = f.read()
       if self._client_key:
           with open(self._client_key, "rb") as f:
               private_key = f.read()

       credentials = grpc.ssl_channel_credentials(
           root_certificates=root_certs,
           private_key=private_key,
           certificate_chain=cert_chain,
       )
       self._channel = grpc.secure_channel(self._grpc_address, credentials)
   else:
       self._channel = grpc.insecure_channel(self._grpc_address)
```
The TLS parameters are read using kwargs.get() rather than kwargs.pop() so they pass through to the REST transport as well. This ensures that when a gRPC-enabled client falls back to REST for non-bulk operations, those REST requests also use TLS with the same certificates.

Certain SSL parameters that have no equivalent in gRPC's Python API remain unsupported and raise NotImplementedError if provided: ssl_assert_fingerprint and ssl_show_warn. When a TLS handshake failure occurs at the gRPC level, the transport detects SSL/TLS-related keywords in the error details and raises SSLError instead of a generic ConnectionError, matching the REST client's behavior so that users' existing exception handling works identically regardless of transport.

### Basic Authentication
Basic auth is implemented via a gRPC UnaryUnaryClientInterceptor that encodes the username and password as a base64 Authorization header and attaches it to the metadata of every outgoing gRPC call. This mirrors how the REST client handles http_auth — where credentials are encoded and added to every HTTP request's Authorization header.
  ```
 class BasicAuthInterceptor(grpc.UnaryUnaryClientInterceptor):
       """gRPC interceptor that adds Basic auth to every unary call."""

       def __init__(self, username: str, password: str) -> None:
           credentials = f"{username}:{password}".encode("utf-8")
           self._auth_header = f"Basic {base64.b64encode(credentials).decode('utf-8')}"

       def intercept_unary_unary(self, continuation, client_call_details, request):
           metadata = list(client_call_details.metadata or [])
           metadata.append(("authorization", self._auth_header))
           new_details = client_call_details._replace(metadata=metadata)
           return continuation(new_details, request)

```
When http_auth is provided to GrpcTransport, the channel is wrapped with this interceptor after creation. Both tuple and string formats are supported, matching the REST client's behavior.

### How JWT works with gRPC
OpenSearch validates JWT tokens server-side using a configured signing key or JWKS endpoint. The client's only responsibility is to pass the token in the Authorization header. For gRPC, this translates to attaching authorization: Bearer as gRPC metadata on every call. The token is self-contained — no client-side signing or cryptography is needed.

The BearerTokenInterceptor accepts tokens with or without the "Bearer " prefix and normalizes them:
```
class BearerTokenInterceptor(grpc.UnaryUnaryClientInterceptor):
       """gRPC interceptor that adds Bearer token auth to every unary call."""

       def __init__(self, token: str) -> None:
           if token.startswith("Bearer "):
               self._auth_header = token
           else:
               self._auth_header = f"Bearer {token}"

       def intercept_unary_unary(self, continuation, client_call_details, request):
           metadata = list(client_call_details.metadata or [])
           metadata.append(("authorization", self._auth_header))
           new_details = client_call_details._replace(metadata=metadata)
           return continuation(new_details, request)
```

The transport detects Bearer tokens in http_auth and creates the correct interceptor:
```
      if isinstance(self._http_auth, (tuple, list)):
          username, password = self._http_auth[0], self._http_auth[1]
          interceptor = BasicAuthInterceptor(username, password)
      elif isinstance(self._http_auth, str) and (
          self._http_auth.startswith("Bearer ")
          or self._http_auth.startswith("bearer ")
      ):
          interceptor = BearerTokenInterceptor(self._http_auth)
      else:
          username, password = str(self._http_auth).split(":", 1)
          interceptor = BasicAuthInterceptor(username, password)
      self._channel = grpc.intercept_channel(self._channel, interceptor)
```

### CI Infrastructure
The Dockerfile is updated to configure OpenSearch with the secure gRPC transport plugin for TLS-enabled integration tests:
   ```
ARG OPENSEARCH_VERSION
   RUN if echo "$OPENSEARCH_VERSION" | grep -qE "^3\."; then \
         echo "aux.transport.types: [secure-transport-grpc]" >> $opensearch_yml \
         && echo "aux.transport.transport-grpc.port: 9400-9500" >> $opensearch_yml \
         && echo "grpc.host: 0.0.0.0" >> $opensearch_yml; \
       fi
```
This uses secure-transport-grpc instead of transport-grpc, enabling TLS on the gRPC endpoint. The gRPC TLS security settings (plugins.security.ssl.aux.*) are passed as Docker container environment variables in run-opensearch.sh, which the default OpenSearch entrypoint writes to opensearch.yml automatically.

### Usage
```
   from opensearchpy.client import OpenSearchGrpc

   # Basic auth with TLS (production)
   client = OpenSearchGrpc(
       hosts=[{"host": "localhost", "port": 9200}],
       grpc_hosts=[{"host": "localhost", "port": 9400}],
       http_auth=("admin", "myStrongPassword123!"),
       use_ssl=True,
       ca_certs="/path/to/root-ca.pem",
   )

   # Mutual TLS (mTLS) with basic auth
   client = OpenSearchGrpc(
       hosts=[{"host": "localhost", "port": 9200}],
       grpc_hosts=[{"host": "localhost", "port": 9400}],
       http_auth=("admin", "myStrongPassword123!"),
       use_ssl=True,
       ca_certs="/path/to/root-ca.pem",
       client_cert="/path/to/client-cert.pem",
       client_key="/path/to/client-key.pem",
   )

   # TLS with system default CAs (no ca_certs needed)
   client = OpenSearchGrpc(
       hosts=[{"host": "localhost", "port": 9200}],
       grpc_hosts=[{"host": "localhost", "port": 9400}],
       http_auth=("admin", "password"),
       use_ssl=True,
   )

   # Basic auth without TLS (development)
   client = OpenSearchGrpc(
       hosts=[{"host": "localhost", "port": 9200}],
       grpc_hosts=[{"host": "localhost", "port": 9400}],
       http_auth=("admin", "admin"),
   )

   # Insecure, no auth — unchanged default behavior
   client = OpenSearchGrpc(
       hosts=[{"host": "localhost", "port": 9200}],
       grpc_hosts=[{"host": "localhost", "port": 9400}],
   )
```


.ci/opensearch/grpc-entrypoint.sh
```diff
@@ -0,0 +1,35 @@
+#!/bin/bash
+# Wrapper entrypoint for gRPC TLS support in CI.
+#
+# Problem: The demo installer checks for "plugins.security" in opensearch.yml.
+# If found, it says "already configured" and skips cert generation.
+# We need gRPC TLS settings (plugins.security.ssl.aux.*) but they must be added
+# AFTER the demo installer runs.
+#
+# Solution: Start a background watcher that adds gRPC TLS settings once the
```
finnegancarroll: This background "watcher" seems pretty brittle. I'm not sure i fully understand the problem here, what happens if we attempt to configure these YML settings as docker container env variables?

opensearch_grpc/grpc_transport.py
```diff
@@ -0,0 +1,304 @@
...
+    - use_ssl=True: Creates a secure gRPC channel (grpc.secure_channel)
+    - ca_certs: Path to CA bundle for server certificate verification
+    - client_cert: Path to client certificate for mTLS
+    - client_key: Path to client private key for mTLS
+
+    When use_ssl=True without ca_certs, system default trusted CAs are used.
+    When use_ssl=False (default), an insecure channel is created.
+
+    Not supported (no gRPC equivalent):
+    - ssl_context: gRPC manages its own SSL internally
```
finnegancarroll: gRPC allows for users to provide an ssl context. It then manages it internally but a user can capture their TLS configuration in an ssl context and we should be able to use their context when initializing our secure channel.

opensearch_grpc/grpc_transport.py
```diff
@@ -0,0 +1,304 @@
...
+    - ssl_version: gRPC negotiates TLS version automatically
```
finnegancarroll: REST will negotiate TLS version as well during ALPN (the initial portion of the TLS handshake where we negotiate protocol/configurations for the connection).

How does REST handle this param? Is this used to force a specific version?

opensearch_grpc/grpc_transport.py
```diff
@@ -0,0 +1,304 @@
...
+    - ssl_assert_hostname: Not configurable in gRPC Python
```
finnegancarroll: I think `ssl_assert_hostname` is `ssl_target_name_override`:
https://grpc.github.io/grpc/core/group__grpc__arg__keys.html#ga218bf55b665134a11baf07ada5980825

I dont see anything for `ssl_assert_fingerprint` yet, might not be supported so sounds good if we throw an exception for this.

finnegancarroll (changes requested):

opensearchpy/client/__init__.py
```diff
@@ -3274,3 +3275,92 @@ def bulk_stream(
...
+
+    # Parameters that are not yet supported for gRPC
+    # (use_ssl, ca_certs, client_cert, client_key, ssl_context, ssl_version,
+    #  ssl_assert_hostname are now supported)
+    _UNSUPPORTED_TLS_ARGS = (
+        "ssl_assert_fingerprint",
+        "ssl_show_warn",
+    )
+
+    _UNSUPPORTED_AUTH_ARGS = ("http_auth",)
```
finnegancarroll: Im wondering if we have a bit of a problem splitting the basic auth and TLS changes into two PRs. That is, in this set of changes how are we authenticating a user on our security enabled test node in CI?

The current tests in this PR should not pass against the CI OpenSearch node because it has the default security plugin demo config which is not ssl_only (FGAC is enabled and users need to authenticate and authorize their operations). Are the current tests silently falling back to the REST path?

Due to this constraint it might make the most sense to include the basic auth changes in this PR as well. That way you can write full end to end integration tests which demonstrate TLS and the various TLS settings, as well as authN/authZ for a basic auth user.

finnegancarroll (changes requested):

opensearch_grpc/grpc_transport.py
```diff
@@ -0,0 +1,391 @@
...
+        self._ssl_assert_hostname = kwargs.get("ssl_assert_hostname", None)
+
+        # Validate single gRPC host — multiple targets not yet supported
+        if self._grpc_hosts and len(self._grpc_hosts) > 1:
+            raise ValueError("Multiple gRPC host targets not yet supported")
+
+        super().__init__(hosts, *args, **kwargs)
+
+        # Resolve gRPC target — grpc_hosts is required
+        if not self._grpc_hosts:
```
harshavamsi, line 286: do we need to validate that only one grpc port can be specified? Not sure if it is already validated elsewhere

opensearch_grpc/grpc_transport.py
```diff
@@ -0,0 +1,391 @@
...
+                except TransportError as e:
+                    if (
+                        hasattr(e, "status_code")
+                        and e.status_code in self.retry_on_status
+                        and attempt < self.max_retries
+                    ):
+                        continue
+                    raise
+
+        return super().perform_request(
```
harshavamsi, line 428: This part looks a bit weird to me, we initialize a gRPC client, and if it isn't a bulk request we use the default transport instead of the gRPC transport. But aren't we missing the client arguments that the default transport supports? Like TLS?

opensearch_grpc/grpc_transport.py
dayana-j: Yes we do because in this version multi grpc channels are not supported. We want to reject multi grpc channels upfront than to silently ignore them, because a client might try to do multi grpc channel and do bulk requests thinking that all the channels they made are running, when only one channel is actually running and the others were silently ignored.

opensearchpy/client/__init__.py
```diff
@@ -3274,3 +3275,92 @@ def bulk_stream(
...
+        # Everything else uses REST
+        client.search(index='my-index', body={'query': {'match_all': {}}})
+
+    :arg hosts: list of REST nodes (same as OpenSearch client).
+    :arg grpc_hosts: list of gRPC nodes, e.g. [{'host': 'localhost', 'port': 9400}].
+    :arg kwargs: all other arguments passed to the OpenSearch client for REST fallback.
+    """
+
+    # Parameters that are not yet supported for gRPC
+    # (use_ssl, ca_certs, client_cert, client_key, ssl_context, ssl_version,
```
finnegancarroll: comment needs updating

.ci/opensearch/grpc-entrypoint.sh
```diff
@@ -0,0 +1,35 @@
...
+# demo installer has created cert files, then call the original entrypoint.
+# The original entrypoint runs the demo installer then starts OpenSearch.
+# OpenSearch takes ~15-20s to start, giving the watcher time to patch the config.
+
+OPENSEARCH_YML="/usr/share/opensearch/config/opensearch.yml"
+
+if [ -f /tmp/.grpc_secure ]; then
+  (
+    # Wait for demo installer to generate certs
+    for i in $(seq 1 60); do
```
finnegancarroll: do we have any way around this "watcher" pattern?

.ci/opensearch/Dockerfile
```diff
@@ -6,3 +6,29 @@ ARG opensearch_yml=$opensearch_path/config/opensearch.yml
...
+      && echo "grpc.host: 0.0.0.0" >> $opensearch_yml; \
+    fi
+
+# Copy wrapper entrypoint (with exec permissions set via --chmod)
+COPY --chmod=755 .ci/opensearch/grpc-entrypoint.sh $opensearch_path/grpc-entrypoint.sh
+
+# Create marker file for 3.x secure mode (needs USER root for /tmp persistence)
+USER root
+ARG OPENSEARCH_VERSION
+ARG SECURE_INTEGRATION
```
finnegancarroll: ARG OPENSEARCH_VERSION
ARG SECURE_INTEGRATION

both duplicates and already declared

opensearchpy/client/__init__.py
```diff
@@ -3274,3 +3275,92 @@ def bulk_stream(
...
+
+    Extends the standard OpenSearch client with gRPC channel management.
+    Bulk requests are routed over gRPC for better performance; all other
+    operations fall through to REST automatically.
+
+    Supported parameters:
+        - hosts: REST endpoint(s) for fallback operations
+        - grpc_hosts: gRPC endpoint (required)
+
+    Unsupported parameters (raise NotImplementedError):
```
finnegancarroll: is this comment out of date?

test_opensearchpy/test_server_secured/test_grpc_secure.py
```diff
@@ -0,0 +1,387 @@
...
+
+    def setUp(self) -> None:
+        if not _grpc_available():
+            self.skipTest(f"gRPC not available on {GRPC_HOST}:{GRPC_PORT}")
+
+    def _get_client(self, **kwargs: Any) -> OpenSearchGrpc:
+        """Create a TLS + auth enabled gRPC client.
+
+        Always uses ca_certs when available since gRPC Python cannot
+        disable certificate verification (unlike REST with verify_certs=False).
```
finnegancarroll, line 64: If the gRPC channel can't configure verify_certs=False, should we just be throwing an error on this case? It doesn't seem like we are handling this in grpc_transport.py. What are we falling back to here?

finnegancarroll (changes requested):
Can we add mTLS integ tests here as well?

finnegancarroll:
Can we rebase on the connection management changes as well when you have a chance. I see some outdated items in the diff which were already addressed in that dependent PR.

harshavamsi (dismissed):

harshavamsi:
@finnegancarroll if this looks good to you I can merge

opensearch_grpc/grpc_transport.py
```diff
@@ -253,33 +475,14 @@ def _raise_grpc_error(self, error: grpc.RpcError) -> None:
...
-            UNAUTHENTICATED (16) → AuthenticationException (401)
-            DATA_LOSS (15)       → TransportError (500)
         """
         code = error.code()
         details = error.details() or "gRPC error"
 
-        if code == grpc.StatusCode.OK:
-            return  # No error
-        elif code == grpc.StatusCode.UNAVAILABLE:
+        if code == grpc.StatusCode.UNAVAILABLE:
```
finnegancarroll, line 541: This looks like an artifact from merging in upstream changes. We want to preserve the full status code mapping here.

opensearch_grpc/grpc_transport.py
```diff
@@ -102,32 +338,22 @@ def perform_request(
...
-
-        If a gRPC handler exists for this request, attempts gRPC first with
-        retries. If gRPC is unavailable after all retries, falls back to REST
-        silently so the operation still succeeds.
-        """
+        """Route to gRPC or REST based on the URL pattern."""
         handler = self._get_grpc_handler(method, url)
         if handler:
-            # Ensure channel is healthy before attempting gRPC
-            self._ensure_channel_connected()
```
finnegancarroll, line 114: Why do we need to remove this?

opensearch_grpc/grpc_transport.py
```diff
@@ -102,32 +338,22 @@ def perform_request(
...
                 except ConnectionTimeout:
                     if self.retry_on_timeout and attempt < self.max_retries:
                         continue
-                    # Fall back to REST on final failure
+                    # Fallback to REST after retries exhausted
                     break
                 except ConnectionError:
                     if attempt < self.max_retries:
-                        # Attempt channel reconnect before next retry
-                        self._reconnect_channel()
```
finnegancarroll, line 128: Wondering about this as well. Can we retry the gRPC request here still? why do we remove this?

I think in general a fatal connection error is something we want to surface to the user instead of silently using REST.

opensearch_grpc/grpc_transport.py
```diff
@@ -86,7 +232,97 @@ def __init__(self, hosts: Any, *args: Any, **kwargs: Any) -> None:
...
+        #   - use_ssl=False: No encryption (insecure channel)
+        if self._use_ssl:
+            # gRPC Python does not support disabling certificate verification.
+            # When verify_certs=False without ca_certs, gRPC will use system CAs
+            # and fail against self-signed certificates. Users MUST provide
+            # ca_certs (or ssl_context with CA loaded) for self-signed certs.
+            if not self._verify_certs and not self._ca_certs and not self._ssl_context:
+                import warnings
+
+                warnings.warn(
```
finnegancarroll: I think this should be a fatal error. Any time a security configuration is not supported for gRPC we should avoid any fallback and just surface the error immediately.

opensearch_grpc/grpc_transport.py
```diff
@@ -50,23 +99,120 @@
     ConnectionTimeout,
     NotFoundError,
     RequestError,
+    SSLError,
     TransportError,
 )
 from opensearchpy.transport import Transport
 
 
+class BasicAuthInterceptor(grpc.UnaryUnaryClientInterceptor):  # type: ignore[misc]
```
finnegancarroll: How do we handle future streaming endpoints? or existing streaming endpoints such as the ML services introduced in https://github.com/opensearch-project/opensearch-py/pull/1096

I think it's fine if we only support these unary operations for now and throw errors for non unary services which are attempting to user one of these auth types.

we can open an issue for this to track future support.

finnegancarroll (changes requested):

finnegancarroll:
Added some minor comments mostly related to upstream changes that came in (rebase with connection management + streaming services).

opensearch_grpc/grpc_transport.py
dayana-j: Right now, it does not support non unary services. The streaming endpoints would skip the inceptor entirely. So, I have went ahead and implemented the intercepter types that handles streaming endpoints for all of the auth interceptor classes, so that the future streaming endpoints are supported.

finnegancarroll (approved):

harshavamsi (approved):

PR #1093, opened 2026-07-14
Add JWT/Bearer token authentication support for gRPC transport

dayana-j:
### Description
This PR adds JWT (JSON Web Token) authentication support for the gRPC transport layer. OpenSearch's security plugin supports JWT authentication over gRPC starting in version 3.5, where the gRPC transport shares the same authentication domains as the HTTP layer. This implementation allows users to pass a pre-obtained JWT token via the http_auth parameter, and the token is attached as a Bearer authorization header in gRPC metadata on every call.

The implementation follows the same interceptor pattern used for Basic auth — a BearerTokenInterceptor that injects the token into gRPC metadata. The transport detects Bearer tokens by checking if the http_auth string starts with "Bearer " and routes to the appropriate interceptor. TLS is required when using JWT over gRPC per the OpenSearch documentation.

### How JWT works with gRPC

OpenSearch validates JWT tokens server-side using a configured signing key or JWKS endpoint. The client's only responsibility is to pass the token in the Authorization header. For gRPC, this translates to attaching authorization: Bearer <token> as gRPC metadata on every call. The token is self-contained — no client-side signing or cryptography is needed.

The BearerTokenInterceptor accepts tokens with or without the "Bearer " prefix and normalizes them:
```
class BearerTokenInterceptor(grpc.UnaryUnaryClientInterceptor):
       """gRPC interceptor that adds Bearer token auth to every unary call."""

       def __init__(self, token: str) -> None:
           if token.startswith("Bearer "):
               self._auth_header = token
           else:
               self._auth_header = f"Bearer {token}"

       def intercept_unary_unary(self, continuation, client_call_details, request):
           metadata = list(client_call_details.metadata or [])
           metadata.append(("authorization", self._auth_header))
           new_details = client_call_details._replace(metadata=metadata)
           return continuation(new_details, request)
```

The transport detects Bearer tokens in http_auth and creates the correct interceptor:

 ```
       if isinstance(self._http_auth, (tuple, list)):
           username, password = self._http_auth[0], self._http_auth[1]
           interceptor = BasicAuthInterceptor(username, password)
       elif isinstance(self._http_auth, str) and (
           self._http_auth.startswith("Bearer ")
           or self._http_auth.startswith("bearer ")
       ):
           interceptor = BearerTokenInterceptor(self._http_auth)
       else:
           username, password = str(self._http_auth).split(":", 1)
           interceptor = BasicAuthInterceptor(username, password)
       self._channel = grpc.intercept_channel(self._channel, interceptor)
```

### Usage

   ```
from opensearchpy.client import OpenSearchGrpc

   # JWT authentication over TLS
   client = OpenSearchGrpc(
       hosts=[{"host": "localhost", "port": 9200}],
       grpc_hosts=[{"host": "localhost", "port": 9400}],
       http_auth="Bearer
   eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJhZG1pbiIsInJvbGVzIjoiYWxsX2FjY2VzcyJ9.signature",
       use_ssl=True,
       ca_certs="/path/to/root-ca.pem",
   )

   # Bulk goes over gRPC with Bearer token in metadata
   resp = client.bulk(body=[
       {"index": {"_index": "my-index", "_id": "1"}},
       {"title": "Authenticated via JWT"},
   ])

   # Search falls back to REST — token is passed via headers
   results = client.search(index="my-index", body={"query": {"match_all": {}}})
   The token can also be passed without the "Bearer " prefix — it gets added automatically:

   client = OpenSearchGrpc(
       hosts=[{"host": "localhost", "port": 9200}],
       grpc_hosts=[{"host": "localhost", "port": 9400}],
       http_auth="eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJzdWIiOiJhZG1pbiJ9.sig",
       use_ssl=True,
   )

```

Basic auth continues to work as before:

   ```
client = OpenSearchGrpc(
       hosts=[{"host": "localhost", "port": 9200}],
       grpc_hosts=[{"host": "localhost", "port": 9400}],
       http_auth=("admin", "myStrongPassword123!"),
       use_ssl=True,
   )
```

### Related

   - OpenSearch JWT Authentication Docs (https://docs.opensearch.org/latest/security/authentication-backends/jwt/)
   - "Using JWT authentication with gRPC" section confirms JWT over gRPC is supported since OpenSearch 3.5, sharing the same authentication domains as the HTTP layer.


finnegancarroll (changes requested):
Can we implement integration tests with JWT over gRPC fully configured in security plugin to validate these changes?

https://docs.opensearch.org/latest/security/authentication-backends/jwt/#using-jwt-authentication-with-grpc

